PR #2, opened 2015-05-27
Broadcast events to the events topic

tie-rack:
* Creates, updates, and deletes are asynchronously sent to the events exchange
* Updated handler functions to take an async channel to send messages to
* Updated elections/update-election return value to include the "before" picture of the election
* Updated tests

Pivotal: [92927786](https://www.pivotaltracker.com/story/show/92927786)


src/election_works/queue.clj
```diff
@@ -24,10 +25,16 @@
...
+    (let [events-exchange "events"
+          events-ch (lch/open @connection)
+          ok-ch (lch/open @connection)
           election-create-ch (lch/open @connection)
           election-update-ch (lch/open @connection)
-          election-delete-ch (lch/open @connection)]
+          election-delete-ch (lch/open @connection)
+          election-create-broadcast-ch (lch/open @connection)
+          election-update-broadcast-ch (lch/open @connection)
+          election-delete-broadcast-ch (lch/open @connection)]
```
cap10morgan: Would `event` be better than `broadcast` here?

src/election_works/queue.clj
```diff
@@ -24,10 +25,16 @@
           (do (Thread/sleep (* attempt 1000))
               (recur (inc attempt)))
           (do (log/error "Connecting to RabbitMQ failed. Bailing.")))))
-    (let [ok-ch (lch/open @connection)
+    (let [events-exchange "events"
+          events-ch (lch/open @connection)
+          ok-ch (lch/open @connection)
```
cap10morgan: What does this do?

src/election_works/queue.clj
```diff
@@ -41,11 +48,16 @@
...
+      (k/responder election-update-ch "election-works.election.update" (handlers/handle-errors (partial handlers/update-election channels/updates)))
+      (k/responder election-delete-ch "election-works.election.delete" (handlers/handle-errors (partial handlers/delete-election channels/deletes)))
+      (k/async->rabbit channels/creates election-create-broadcast-ch events-exchange "election.create")
+      (k/async->rabbit channels/updates election-update-broadcast-ch events-exchange "election.update")
+      (k/async->rabbit channels/deletes election-delete-broadcast-ch events-exchange "election.delete")
       {:connections [@connection]
-       :channels [ok-ch election-create-ch election-update-ch election-delete-ch]})))
+       :channels [ok-ch election-create-ch election-update-ch election-delete-ch
+                  election-create-broadcast-ch election-update-broadcast-ch election-delete-broadcast-ch
+                  events-ch]})))
```
cap10morgan: Should these be sets instead of vectors?

src/election_works/channels.clj
```diff
@@ -1,2 +1,10 @@
 (ns election-works.channels
   (:require [clojure.core.async :as async]))
+
+(def creates (async/chan))
+(def updates (async/chan))
+(def deletes (async/chan))
```
cap10morgan: Would it be better to name these `create-events`, for example, to make it clearer that they are for broadcasting events after the fact rather than for the incoming commands?

src/election_works/queue.clj
tank157: In the case of channels, we did order them so that event-ch is the last, so that the broadcast channels are closed first. So there was an intentional ordering there, though whether or not that really matters in the end is somewhat up in the air.

src/election_works/queue.clj
tank157: The ok-ch? It's in the kraken lein template, and I assume it's there to act in much the same way the /ping http endpoint.

src/election_works/queue.clj
tank157: *Shrug* I guess...neither seem particularly bad, and the scope of the var name is just this fn, but I can see event being better.

src/election_works/queue.clj
cap10morgan: OK. If you want order to be significant, then :+1:. The closing code should all be using the ISeq interface and shouldn't care if it's a set or a vector.

src/election_works/channels.clj
tank157: Well, if we make the change to the rabbit channel to `election-create-events-ch` as above, and the async channel is `create-events`, there's both symmetry there but also the potential for confusing the two? At least the rabbit ch symbol lives only inside the one function.

cap10morgan: I vote symmetry. :)

src/election_works/queue.clj
cap10morgan: Yeah, OK. If that is the case can we call it `ping` here (and eventually in the template, but that's out this PR's scope)? `ok` makes me think it's a success response channel or something like that.

cap10morgan: Sorry, I realize this being "touched by the PR" is a technicality. Feel free to ignore my comments on it.

tank157: Yeah, let's have a wider team discussion on this. I definitely like the idea behind it, but the name could be under discussion. I'd like `heartbeat` myself.

tie-rack: `heartbeat` suggests that it's sending messages at a regular interval on its own... though perhaps a heartbeat topic would be very useful for monitoring.

src/election_works/channels.clj
tie-rack: I'll update for symmetry.

tie-rack:
@cap10morgan: Names updated. How's it look now?

cap10morgan:
LGTM

PR #3, opened 2015-06-08
Add :election/id UUID lookup

cap10morgan:
[Pivotal card](https://www.pivotaltracker.com/story/show/95884842)

The immediate need for this change is to support election notifications. The to-be-created electorate-works component will be storing (election-id, user-id, election-authority-id, registration-authority-id) tuples, so it will need to be able to retrieve the full representation of each of these by their UUIDs. This PR adds that capability for election-id's.

The dependency updates were purely opportunistic and didn't seem to break anything.


ericnormand:
I like the standardization effort!

Questions I have seeing this pull request:

* Why is this change being made?
* Are the updated deps necessary for this change (or were they convenient and unrelated)?

What do you think about answering these questions in the PR description so that it becomes part of the commit log?

ericnormand:
I ran the tests with `lein test` and it all passed.

src/election_works/handlers.clj
```diff
@@ -28,10 +29,21 @@
...
+  "A handler for reading elections"
+  [{:keys [election-id] :as message}]
+  (log/debug "Election read handler received message:" message)
+  (or (some->> election-id
+               (e.election/find-by-id (db/db))
+               e.election/entity->election
+               (assoc {:status :ok} :election))
+      {:status :not-found
+       :election-id election-id}))
+
```
ericnormand: Does kehaar do error handling? What if this throws an exception? Where does the exception get caught and logged? Does a response message get sent?

ericnormand:
Test coverage does indicate the that the found case is being tested. What do you think about adding a case where the election does not exist? And what about a case where the message is malformed?

src/election_works/handlers.clj
tie-rack: Kehaar's not totally there yet. It does need to be able to handle uncaught exceptions from handler fns. But that's outside of this PR's scope.

cap10morgan: I agree with @tie-rack, but I should point out that there is [some error-handling built in to election-works](https://github.com/democracyworks/election-works/blob/master/src/election_works/handlers.clj#L12-L29) that [this takes advantage of](https://github.com/democracyworks/election-works/blob/add-uuid-lookup/src/election_works/queue.clj#L61-L62). Could be a pattern to extract into kehaar at some point.

cap10morgan:
@ericnormand Good call. I'll see about adding some tests for those cases.

src/election_works/handlers.clj
ericnormand: I see now that errors are handled. Thanks!

ericnormand:
One blocker: test cases for not-found case and malformed message case.

After that, LGTM!

cap10morgan:
@ericnormand Adding the malformed message test drove out an improvement to the code. You may want to take a look at that last commit.

src/election_works/entities/election.clj
```diff
@@ -50,6 +50,14 @@
                [denamespaced-key (map ocd/entity->ocd value)]
                [denamespaced-key value]))))
 
+(def read-schema
+  {:election-id s/Uuid})
+
+(defn read-election
+  [params]
+  (s/validate read-schema params)
+  ((comp entity->election find-by-id) (db/db) (:election-id params)))
```
ericnormand: Is this the clearest way to write this? Why not threading or simply `(entity->election (find-by-id ...))`?

ericnormand: I suggest we follow the convention of nil meaning not found and non-nil meaning found. It is not obvious that an empty map means not found.

src/election_works/handlers.clj
```diff
@@ -28,10 +28,19 @@
          :error {:type :server
                  :server (str "Unknown server error: " (.getMessage runtime-exception))}}))))
 
+(defn read-election
+  "A handler for reading elections"
+  [message]
+  (log/debug "Election read handler received message:" message)
+  (let [election (e.election/read-election message)]
+    (if (empty? election)
+      (merge {:status :not-found} message)
```
ericnormand: What do you think about putting the original message in a sub map? For example: `{:status :not-found :request {:election-id ...}}`. It seems like a clearer convention to me.

ericnormand:
Tests run with `lein test` again.

src/election_works/handlers.clj
cap10morgan: I have no objection to that per se, but the convention I'm copying here comes from user-works where it puts the id you were searching for in the not-found return (with its key). Since the schema enforces that that is all that is in `message`, there wasn't anything else to do to it.

But if you'd rather I pull out the id explicitly anyway as a future-proofing move, I could get behind that.

cap10morgan:
@ericnormand Ready for your review again. Thanks for the good suggestions!

src/election_works/handlers.clj
ericnormand: It sounds like this is a broader discussion than this PR. I'll bring it up in Slack.

> But if you'd rather I pull out the id explicitly anyway as a future-proofing move, I could get behind that.

No, merging the whole thing is better.

src/election_works/entities/election.clj
```diff
@@ -50,6 +50,17 @@
...
+(def read-schema
+  {:election-id s/Uuid})
+
+(defn read-election
+  [params]
+  (some->> params
+           (s/validate read-schema)
+           :election-id
+           (find-by-id (db/db))
+           entity->election))
```
ericnormand: NOT A BLOCKER, JUST DISCUSSION

I'm not quite convinced that pulling params out of a map is the right level of abstraction for this function. What do you think about it taking just the ID and having the handler (where the message enters the system) do the validation?

src/election_works/handlers.clj
```diff
@@ -28,10 +28,18 @@
...
                  :server (str "Unknown server error: " (.getMessage runtime-exception))}}))))
 
+(defn read-election
+  "A handler for reading elections"
+  [message]
+  (log/debug "Election read handler received message:" message)
+  (if-let [election (e.election/read-election message)]
+    (assoc {:status :ok} :election election)
+    (merge {:status :not-found} message)))
+
```
ericnormand: This is really clear now! :)

ericnormand:
Tests pass with `lein test`.

ericnormand:
LGTM

src/election_works/entities/election.clj
cap10morgan: Hmm... I have two thoughts on that:

1. This is how all of the counterpart fns (create, update, & delete) work. So the scope of this card seemed constrained to follow that convention. That, of course, doesn't we can't or shouldn't change it.
1. I like that the schema is checked in the entity because then it doesn't matter where the params are coming from, they always have to conform to the schema.

ericnormand: 1. Totally out of scope with the card, so no problem.
2. I don't like checking the schema there for the same reason: it should not matter where it came from. But in addition, it should not matter what the message format is. That's someone else's concern.

The only thing that should concern this function is that if I have a UUID, I should be able to look up an election.

Hypothetical example:  
If were were going to add an HTTP GET API for lookups, we would want the uuid to be a segment in the path. It would not conform to this schema. In order to do the lookup, we would need to both convert the string from the path into a UUID (necessary anyway) and box it up in a map (not necessary).

cap10morgan: Right, I agree with you! But I think we're not *trying* to enforce the message schema here. Instead, because of which fn this validation is attached to, we're enforcing this fn's schema. Just like we do with create, update, and delete.

However, now that you've got me thinking about it, I do think it would be better if this fn just took an id. It could still validate that. Maybe I'll send up a change that does that...

PR #4, opened 2015-06-09
Timeout in tests when checking events chans

ericnormand:
We need to test that the proper events are added to the event queues. We were using blocking `async/<!!` to take from the chan. If the code under test *does not* add a message, then this will block forever, stopping all further tests from running, and never reporting the source of the problem.

The solution I propose is to timeout on the take, using `clojure.core.async/alt!!`, with a timeout in one of the branches. I chose one second as the timeout. It seems like a reasonable upper bound for writing to a local datomic, yet still low enough for fast failures. If the timeout is triggered, it prints a message indicating the lack of event on the event channel.


ericnormand:
Assigning to @tie-rack.

tie-rack:
Agreed. LGTM!

PR #5, opened 2015-06-09
Misc cleanups

cap10morgan:
Some of the other improvements we discussed in Slack.

1. Use `:id` instead of `:election-id` to be consistent w/ user-works (and we decided in #dev that we preferred its approach to the old one in here).
1. Make the read and delete entity fns just take id's. There's no reason they need to take single-element maps. As a consequence, the incoming read and delete queue messages now have different schemas, so those are checked separately in the `handlers` ns.


src/election_works/entities/election.clj
```diff
@@ -70,15 +68,15 @@
 (defn create-election
   [params]
   (s/validate create-schema params)
-  (if-let [election-id (:election-id params)]
-    (if-let [election (find-by-id (db/db) election-id)]
+  (if-let [id (:id params)]
+    (if-let [election (find-by-id (db/db) id)]
```
ericnormand: Since there's a repeated line `(entity->election ...`, what do you say to collapsing this into a single `if`?

```
(if (and (:id params) (find-by-id (db/db) (:id params)))
  (throw ...)
  (entity->election ...))
```

cap10morgan: Heh, not technically touched by this PR, but since the theme here is "misc cleanups" I guess everything's on the table! New commit incoming...

src/election_works/handlers.clj
```diff
@@ -62,11 +71,14 @@
...
 (defn delete-election
   "A handler for deleting elections"
-  [deletes-channel {:keys [election-id] :as message}]
+  [deletes-channel {:keys [id] :as message}]
   (log/debug "Election delete handler received message: " message)
-  (do (e.election/delete-election
-       {:election-id election-id})
-      (async/put! deletes-channel election-id)
+  (s/validate delete-schema message)
+  (do (e.election/delete-election id)
```
ericnormand: Why are you using do here?

ericnormand:
LGTM

src/election_works/handlers.clj
cap10morgan: *I'm* not, it was just already there. But you're right, it can be gotten rid of now.

PR #6, opened 2015-06-10
Search election by ocd-ids

ericnormand:
Create a new queue and handler for searching for upcoming elections by
ocd-id. We can send a message containing one or more ocd-ids to this
queue and expect a response of matching upcoming elections.

At least one ocd-id must match exactly. "Upcoming" is defined as >=
today at the beginning of the day. We validate input in the message
handler and in the entity function that performs the search. It returns
a list of found elections in the same format as returned by read.

[Pivotal Card](https://www.pivotaltracker.com/story/show/95877590)


ericnormand:
Assigned to @tank157

ericnormand:
`lein test` passes.

Also we manually tested the wiring of the new queue.

src/election_works/entities/election.clj
```diff
@@ -59,6 +61,28 @@
...
 
+(def search-by-ocd-id-schema
+  [(s/one division-regex "At least one division id.") division-regex])
+
+(defn search-elections-by-ocd-id
+  [ids]
+  (s/validate search-by-ocd-id-schema ids)
+  (let [db (db/db)
+        date (time.coerce/to-date (time/today))]
+    (->> (d/q '[:find ?election
```
tank157: Chris has been pushing for using the map literal form of queries, but that's just a nice to have. Admittedly, I tried to convert one over once, and couldn't easily find docs on the map literal form so take of that what you will.

tank157:
Other than that totally optional query format, LGTM. Nice work!

cap10morgan:
I reviewed this too and it L(Great)TM.

src/election_works/entities/election.clj
ericnormand: Thanks for letting me know. I think we should keep it consistent with the rest of the code in this file and make another card to translate the format as a whole.

PR #7, opened 2015-06-12
[TH] change config from ballot-notification-works to election-works

tank157:
Should be connecting to the election-works database, not ballot-notification-works.

No card, just noticed while looking at stuff.

Assigning to @bwreid


bwreid:
Oh yeah we should fix that. Thanks Troy! LGTM.

PR #8, opened 2015-06-18
Authority level

ericnormand:
Related to [this Pivotal Card](https://www.pivotaltracker.com/story/show/95886418)

We need to store what level of authority (state, county, municipal) is responsible for voting at a particular election. This needs to be set explicitly at election creation (we can't calculate it). This will be used to look up which election authority a user should use to vote.

For example, a state election (governor) might have the authority-level set to :municipal. In this case, we look up the municipal OCD-ID for the user and find the authority that corresponds.

This change merely stores the authority-level. It does not do that search (which happens in electorate-works).

The authority-level is based on https://torchpad.com/workspace/wikis/democracyworks/pages/data+models/elections

> election/authority-level (keyword; one)
will be one of :state, :county, or :municipal
defines which of the users' registered address OCD-IDs should be used to match up with an election authority for vote-by-mail and all other purposes (sometimes these are different from the authority handling voter reg.)


ericnormand:
Assigning to @tank157.

ericnormand:
`lein test` passes; I did not test the queues locally since I only changed message formats.

tank157:
Need to also add `election/voter-registration-authority-level (keyword; one)` attribute.

tank157:
LGTM

PR #9, opened 2015-06-18
[TH] legacy initialize needs to go

tank157:
No Pivotal Card, just something I noticed. It gets called a few lines later.

Assigning to @cap10morgan


cap10morgan:
LGTM. And I would say that is falls into the realm of "feel free to just fix and push to master."

PR #13, opened 2015-07-06
exclude nop and log4j libs from datomic & add logback

cap10morgan:
Without this I got nop logging (i.e. no log output) when I combined this
with another project in a docker-compose stack.

No card, just an opportunistic improvement.


ericnormand:
LGTM.

ericnormand:
How'd you figure that out?

cap10morgan:
Oh, the significant difference here was not using WildFly. The Immutant docs says that outside of WildFly, it uses logback for logging, but inside WildFly it uses WildFly's logging subsystem. So that makes sense.

cap10morgan:
@ericnormand [Datomic logging docs](http://docs.datomic.com/configuring-logging.html)

PR #15, opened 2015-07-28
Store deadlines for elections

ericnormand:
We add two types of election options to the [election data model]: voting-option and voter-registration-option.  Currently the options only store dates and deadlines. There are more pieces of data that these will store, but that was not part of this task.

[election data model]: https://torchpad.com/workspace/wikis/democracyworks/pages/data+models/elections

[Pivotal card](https://www.pivotaltracker.com/story/show/98572924)

We modified create and update for elections. There are tests in handlers-test and elections-test.

We've also identified a potential race condition which is similar to other conditions we are ignoring for the demo. We've made a pivotal card to address that issue [here](https://www.pivotaltracker.com/story/show/100038256).

More information is in the commit messages.


resources/schemas/20150723deadlines.edn
```diff
@@ -0,0 +1,149 @@
+[
```
cap10morgan: Can we rename this file to something like `20150723-add-voting-option-voter-registration-option-and-deadline.edn`?

resources/schemas/20150723deadlines.edn
```diff
@@ -0,0 +1,149 @@
...
+  :db.install/_attribute :db.part/db}
+
+ {:db/id #db/id[:db.part/db]
+  :db/ident :voter-registration-option.by-mail/deadline.in-person
+  :db/valueType :db.type/ref
+  :db/cardinality :db.cardinality/one
+  :db/doc "Deadline for returning a ballot by-mail (in-person
+  date). Used with :by-mail type."
+  :db/isComponent true
+  :db.install/_attribute :db.part/db}
```
cap10morgan: The data model doc also calls for a `:voter-registration-option/authority` UUID attr and a `:voter-registration-option/note` string attr.

resources/schemas/20150723deadlines.edn
```diff
@@ -0,0 +1,149 @@
...
+  :db/doc "The state whose deadline we're overriding. This exists
+  primarily for supporting national election deadline overrides"
+  :db.install/_attribute :db.part/db}
+
+ {:db/id #db/id[:db.part/db]
+  :db/ident :deadline/date
+  :db/valueType :db.type/instant
+  :db/cardinality :db.cardinality/one
+  :db/doc "The date (and optional time) of this deadline."
+  :db.install/_attribute :db.part/db}
```
cap10morgan: `:deadline/note`?

src/election_works/entities/election.clj
```diff
@@ -48,14 +174,117 @@
...
+       (for [voting-option voting-options]
+         [:db/retract id :election/voting-option (:db/id voting-option)]))
+     (when (contains? params :voter-registration-options)
+       (for [voter-registration-option voter-registration-options]
+         [:db/retract id :election/voter-registration-option (:db/id voter-registration-option)])))))
+
+(defn update-transact-election
+  [old-election params]
+  (let [tx-data (->tx-data params)
+        retractions (->retractions old-election params)
```
cap10morgan: What happens if you just assert a new set of refs on a cardinality many ref attribute without retracting each of the old set first? Does Datomic union the new w/ the existing?

src/election_works/entities/election.clj
```diff
@@ -24,22 +26,146 @@
...
+                 {:db/id                                                 ?id
+                  :voter-registration-option/type                        type
+                  :voter-registration-option.in-person/start             start
+                  :voter-registration-option.in-person/end               end
+                  :voter-registration-option.by-mail/deadline.received   (first deadline-received-tempids)
+                  :voter-registration-option.by-mail/deadline.postmarked (first deadline-postmarked-tempids)
+                  :voter-registration-option.by-mail/deadline.in-person  (first deadline-in-person-tempids)}
+                 [:voter-registration-option.in-person/start :voter-registration-option.in-person/end
+                  :voter-registration-option.by-mail/deadline.received :voter-registration-option.by-mail/deadline.postmarked
+                  :voter-registration-option.by-mail/deadline.in-person])]))})
```
cap10morgan: Would using [nested maps in transactions](http://docs.datomic.com/transactions.html#sec-3-5) make this simpler?

test/election_works/test_helpers.clj
```diff
@@ -3,9 +3,10 @@
             [datomic-toolbox :as db]
             [turbovote.resource-config :refer [config]]))
 
-(def fresh-db
-  (do (db/initialize)
-      (d/db (db/connection))))
+(comment
+ (def fresh-db
+   (do (db/initialize)
+       (d/db (db/connection)))))
```
cap10morgan: Should we just delete this code?

src/election_works/entities/election.clj
ericnormand: Yes. It only adds, even if they are identical maps (because they become new entities with different db/ids).

test/election_works/test_helpers.clj
ericnormand: Yes. Thanks.

src/election_works/entities/election.clj
ericnormand: Possibly. I'm looking into it.

ericnormand:
I think we addressed everything.

I made the note property common among everything. It's called :admin/note since they seemed to be used for the same purpose (basically a comment).

ericnormand:
Ping @cap10morgan, thanks for reviewing. It's getting better.

ericnormand:
db.fn/retractEntity retracts all datoms where the given id is in the entity position and all datoms where the id is in the value position. http://docs.datomic.com/transactions.html#outline-container-4-1

ericnormand:
We can make notes optional.

cap10morgan:
There are a couple of outstanding questions about the data model that we'd like to figure out before merging and deploying this. I'll try to get those answered tomorrow (7/30).

cap10morgan:
This is hopefully ready to move forward again. I should probably pair with someone on it to bring in the latest data model changes.

ericnormand:
Ping @cap10morgan . Data model changes are live.

src/election_works/entities/election.clj
```diff
@@ -48,15 +146,121 @@
         db-after (:db-after @(db/transact tx-data))]
     (find-by-id db-after id)))
 
+(defn ->retractions [election-entity params]
```
tank157: Might be nice to either rename this to include district-divisions in the name, to be clearer about what is being retracted.

tank157:
Minor comments aside, and assuming the schema comments will also be handled properly, there's still one big issue here. Namely, without changing electorate-works, these changes will break functionality. The relocation of the authority levels to district-divisions, and the need to match a district division to the user to find the proper authority levels means that until we fix those, we will possibly cause exceptions at worst, or not matching authorities at best. This may be fine, as the upcoming election notification doesn't currently use authority data, but the vote tomorrow one might, and we will need to fix it soon.

In lieu of a PR against electorate-works, I'd be fine knowing that this was tested in krakenstein, that it doesn't cause errors that break the connections in kehaar, and a card is written to fix this. And I'm happy to help try this out in krakenstein.

src/election_works/entities/election.clj
ericnormand: It is all retractions related to an election, so it follows the same pattern as ->tx-data

ericnormand:
OK, I'm making a card for electorate-works.

tank157:
Presuming the merge will be handled with care, and corrdinated with electorate-works as per the other card, LGTM

ericnormand:
Made a card here. https://www.pivotaltracker.com/story/show/100578620

Don't merge except as a part of that card. Thanks!

PR #16, opened 2015-08-14
Early lein immutant war to avoid dep downloads

ericnormand:


tie-rack:
I think you can do direct to master for these. It looks _that_ good to me.

PR #18, opened 2015-09-28
add election-day voter reg. to schema

cap10morgan:
We're about to import some elections w/ `:election-day` voter-registration-options. This allows them in the schema.


ericnormand:
LGTM

PR #19, opened 2015-09-30
Election import

cap10morgan:
I used this code to import the Nov. 3 elections from Carmen's Google Sheets.


cap10morgan:
I need to back out that last commit, it shouldn't be on this branch.

dev-src/election_works/import.clj
```diff
@@ -0,0 +1,101 @@
...
+                                                   (row-get "deadline:reg-by-mail")
+                                                   clojure.instant/read-instant-date)}}
+                                            (#(if (= (row-get row "election-day-registration")
+                                                     "TRUE")
+                                                (conj % {:type :election-day})
+                                                %)))}))))
+
+(defn row->election
+  [row]
+  (-> {:description (row-get row "description")
```
cap10morgan: This threading macro should not be here.

However, we're using it #21 which will get merged as soon as this one is LGTM'd. OK to overlook for now on those grounds?

dev-src/election_works/import.clj
```diff
@@ -0,0 +1,101 @@
...
+                                                   (row-get "registration-deadline-type")
+                                                   (#(str "deadline-" %))
+                                                   keyword)
+                                               (-> row
+                                                   (row-get "deadline:reg-by-mail")
+                                                   clojure.instant/read-instant-date)}}
+                                            (#(if (= (row-get row "election-day-registration")
+                                                     "TRUE")
+                                                (conj % {:type :election-day})
+                                                %)))}))))
```
tie-rack: Please make that logic its own function. `(#(..))` in a threading macro is not very readable.

tie-rack: There are many other functions that seem to be begging to be broken out as well.

cap10morgan:
@tie-rack Thanks for the review. I extracted several functions and added a test of the import code.

tie-rack:
What do you say about turning `println` and `prn` into logging statements?

cap10morgan:
@tie-rack What would that get us in code explicitly intended to be run at the CLI?

tie-rack:
Nothing _really_. `println` isn't atomic or thread-safe and _can_ result in intertwined messages. I suppose it's unlikely here, but as a way to log what's happening, there's a better option already included in this project.

README.md
```diff
@@ -114,6 +114,24 @@ The response will include the fully updated election.
...
+
+`lein run -m election-works.import '{:host "rabbit-host" :port rabbit-port}' /path/to/elections.csv`
+
+The CSV file should have the following headers (value hints in parens):
+
+* date (election date)
+* description (description of election)
+* registration-authority (one of "state", "county", or "municipal")
+* vote-by-mail-authority (one of "state", "county", or "municipal")
+* district-division (space-separated list of OCD Division IDs defining who can vote in this election)
```
tank157: Should we pluralize district-division here and in the code? Would seem to make sense with the data it expects.

cap10morgan: It started out singular in the input spreadsheets, so I kept that here. But sure, I can see if Carmen can make that change everywhere.

cap10morgan:
@tie-rack Yeah, we've run into that before in mail-batcher (concurrent println's stepping all over each others' messages) and wrapped it in a core.async channel to fix it.

The only reason I'm hesitant to use logging in situations like this is that it's very easy to get into a configuration state where the log messages are printed nowhere.

But I don't feel strongly one way or the other, so if you still do, I will defer to your judgment.

dev-src/election_works/import.clj
```diff
@@ -0,0 +1,114 @@
...
+        rabbit-conn (kehaar.rabbitmq/connect-with-retries rabbit-config)
+        rabbit-ch (wire-up/external-service rabbit-conn
+                                            ""
+                                            "election-works.election.create"
+                                            {:exclusive false
+                                             :durable true
+                                             :auto-delete false}
+                                            5000
+                                            import-chan)
+        import-row (comp async/<!! import-election log-election row->election)]
```
tank157: Is this going to do anything if the backend rejects an election for some reason? I suppose we'll see it in the prn above? Should we do anything more substantive? I'm fine with "no", just wondering.

cap10morgan: Yeah, for now I've just been searching the output for "error". I wasn't planning on doing anything more at this stage, but I'm open to suggestions.

tank157: Since this is still a pretty manual tool, what I'd suggest is perhaps storing the errors aside and then presenting them all at the end, so whomever has to deal with them doesn't have to scroll through all the successful output and search for them. That might be useful.

cap10morgan:
@tank157 It prints this now at the end (w/ an example error row):

```clojure
Results:
{:successful 9,
 :errors 1,
 :error-rows
 [{:result
   {:status :error,
    :error
    {:type :validation,
     :message
     "Value does not match schema: {:district-divisions #{{:voter-registration-options #{(not (some (check % a-clojure.lang.PersistentArrayMap) schemas))}}}}"}},
   :row
   {"vote-by-mail-authority" "county",
    "deadline:reg-by-mail" "2015-10-05",
    "election-day-registration" "FALSE",
    "registration-authority" "county",
    "registration-deadline-type" "postma",
    "district-division"
    "ocd-division/country:us/state:az/place:flagstaff ",
    "type" "Special",
    "date" "2015-11-03",
    "deadline:ballot-request" "2015-10-23",
    "description" "City of Flagstaff Special Election"}}]}
```

cap10morgan:
OK @tank157 said if I got the original row into the error report this would be good. So I'm merging.

tank157:
Yeah, I was just looking for the row number, but this should suffice as well

PR #20, opened 2015-09-30
make datomic conn / db args in entity ns

cap10morgan:
This makes it *much* easier to use the election entity namespace from a
REPL connected to the production Datomic transactor. I needed to do this
to find and delete an election that was cancelled.

I think in general we should try to limit our usage of the stateful
datomic-toolbox fns and instead take datomic connections and db's as
args to functions as much as possible.


src/election_works/entities/election.clj
```diff
@@ -335,12 +335,12 @@
...
 (defn create-election
-  [params]
+  [datomic-conn db params]
   (s/validate create-schema params)
   (let [id (:id params)]
-    (if (and id (find-by-id (db/db) id))
+    (if (and id (find-by-id db id))
       (throw (IllegalArgumentException. "Election with id already exists"))
-      (entity->election (transact-election params)))))
+      (entity->election (transact-election datomic-conn params)))))
```
ericnormand: Just a note (that is, this is fine for now): this is a non-atomic operation. The telltale sign is needing both the current value of the db and a connection to it :)

ericnormand:
Looks good to me!! LGTM!!!

ericnormand:
Plus I agree with the notion that we should be passing in the db/connection.

For reads, pass in the db.

For updates, pass in the connection. If it needs to read *and* update (which is often the case), we should use `datomic-toolbox.transaction/swap-tx!` (which takes just the connection).

cap10morgan:
@eric Agreed on the need for swap-tx! We need that in a released version of datomic-toolbox. And I believe *that* is currently blocked by some design questions about my namespace reorganization PR. Though I suppose it doesn't have to be.

PR #21, opened 2015-09-30
Add Google election IDs

cap10morgan:
[Pivotal card](https://www.pivotaltracker.com/story/show/104444878)

Based on the election-import branch since the card calls for accepting google-election-ids from imports and that's still in a separate PR.

Most of this should be pretty straightforward except for commit 88e06a6. When we went to add `:election/google-election-id` to the optional values that should be dissoc'd from tx-data when nil, we noticed that several required attributes were also being dissoc'd when nil, which isn't right. Then we realized that this was because the update fn was using the same `->tx-data` and on updates, most attrs are optional because you don't necessarily want to change all of them. If you, dear reviewer, feel that this commit belongs in its own, separate PR, we will happily move it into one. Just say the word.


src/election_works/entities/election.clj
```diff
@@ -135,13 +136,42 @@
...
+       date
+
+       :election/district-division
+       (map election-district-division->tx-map
+            district-divisions)
+
+       :election/google-election-id
+       google-election-id}
+      [:election/description :election/date :election/district-division
+       :election/google-election-id])]))
```
tie-rack: The distance between the call to `db/dissoc-optional-nil-values` and it's optional nil values is nearly 20 lines. For readability's sake, I'd suggest `let`ting that map (as `tx-data`, perhaps) and use that as the first arg to `db/dissoc-optional-nil-values`.

cap10morgan:
@tie-rack That better?

tie-rack:
Well now you've got weird whitespace, but yes.

cap10morgan:
@tie-rack Whoops, yep. Fixed now. Anything else?

tie-rack:
This looks good.

cap10morgan:
@tie-rack I think this needs another (hopefully quick) look.

tie-rack:
This thing keeps moving from under you, doesn't it? More conflicts.

cap10morgan:
@tie-rack Oof, sure does. OK, another merge coming.

cap10morgan:
@tie-rack OK ready again (for now).

tie-rack:
The election-id parts look fine... Though this seems to be behind the `import` branch its based off of. I guess this PR has to wait for approval for #19.

PR #22, opened 2015-10-07
Rename voting option

ericnormand:
This branch renames `:voting-option` to `:voting-method` and `:voter-registration-option` to `:voter-registration-method`. This was done for consistency.

We chose to do the versioned queue approach. We've made v2 queues where appropriate for handling the new message format. The old handlers now transform their message to new messages and pass them to the new handlers.

Datomic schemas are also changed using schema alterations. Which are really nice, both the old name and the new name are available. Since it's just a name change, that's good.

Using the versioned queue approach means we can deploy it without worrying about rollover problems.

[Pivotal Card](https://www.pivotaltracker.com/story/show/103457276)

Please check the queue wirings. There are no tests for those.


ericnormand:
Assigned to good old reliable @tank157

README.md
```diff
@@ -27,9 +27,15 @@ Keys in Consul should be appropriately namespaced, preferably under election-wor
 
 ## Usage
 
-Election Works exposes three Rabbit queues: "election-works.election.create", "election-works.election.update", and "election-works.election.delete". Each one will respond on a provided `reply_to` queue with the provided `corellation_id` for each message.
+Election Works exposes three Rabbit queues:
```
tie-rack: I count five queues.

tie-rack:
No tests for v2?

ericnormand:
There are.

tie-rack:
Oh. I see. Looks good to me.

PR #24, opened 2015-10-09
Add instructions to voting-methods and voter-registration-methods

ericnormand:
Store in datomic (new schema) and test create, update, read that it works with new data.

[Pivotal Card](https://www.pivotaltracker.com/story/show/103449296)


bwreid:
LGTM. :+1:

PR #25, opened 2015-11-03
update to docker-toolbox 2.0.0

cap10morgan:
This is just the minimal changes needed to keep things working with the new version. It doesn't make use of swap-tx!.

I did this to help with some REPL work I was doing for vote-tomorrow notifications.

[Pivotal card](https://www.pivotaltracker.com/story/show/106511302)


ericnormand:
Tested with `lein run` and ran it with `docker-compose`. 

LGTM

PR #26, opened 2015-11-03
import state authority defaults w/ elections

cap10morgan:
Someday this will be handled by an elections admin tool. For now, let's just import the state's voting-methods and voter-registration-methods so that elections have the basic data we expect them to.

I did this in order to get voting ID instructions from the state authorities in election-authority-works into elections for the vote-tomorrow notifications that just went out. I used this code in a REPL to do that as the elections were already imported.

But I thought since I was writing code to pull this stuff from election-authority-works anyway and we'll need this again for any elections we import between now and launching an admin tool that can do this, it seemed better to bring it into the importer now.

I tested it with a new election I imported for Denver (it needed a different OCD-ID) and it pulled in the state defaults from the authority.


tie-rack:
Yeah... This looks okay.

PR #27, opened 2015-11-13
log put! results when broadcasting events

cap10morgan:
This is one small part of the effort to figure out why Brooklyn folks don't have electorates. This will log when / if we broadcast election create events. We are put!'ing them which means the code moves on before they are actually sent and if they don't send for some reason we wouldn't know it.

I added the logback.xml so I could confirm that I was actually logging what I thought I was.

No card; Friday work.


7maples:
LGTM

PR #28, opened 2015-12-28
Remove v1 election format

cap10morgan:
We were still using the deprecated queue name versioning strategy in here, but very little was depending on it.

I was about to start on a new election-http-api for finding upcoming elections for users, but rather than coding `v2` queue names into another component (and piling on the technical debt), I thought it would be better to just remove the barely-needed v1 support here first instead.


tank157:
LGTM

PR #29, opened 2016-01-11
No search results success

cap10morgan:
Bifrost expects searches w/ no results to be empty success responses, not errors (o/w you get 500s when you're expecting empty 200s).

This PR fixes that in election-works so that election-http-api works as expected.


danielglauser:
LGTM

PR #32, opened 2016-01-22
push latest and $DEPLOY_ENV docker tags

cap10morgan:
Now that I've tested this on a couple of projects and it seems to be working well, it's time for a PR review.

So here's what this does and the thinking behind it:

1. Every successful image build that gets pushed to Quay with the `:branch-sha` tag will *also* have a `:latest` tag assigned and pushed to Quay. This tag is what gets used if you, for example, run `docker pull quay.io/democracyworks/project` (and don't specify a tag). It seems appropriate here b/c it will indeed always be the latest image we've pushed.
1. On deploy, we specify the deploy environment (`staging` or `production`) as an arg to the deploy script in BuildKite and–on successful deploy–that gets assigned as a new tag to the image and pushed to Quay as well.
1. This then allows us to say in the krakenstein docker-compose.yml: `image: quay.io/democracyworks/project:production` and use the image running in production in krakenstein.

This is part of [this card](https://www.pivotaltracker.com/story/show/112161883) and a necessary component of using images by default in krakenstein rather than building local working copies.

Other changes that come in are a result of copying the build and deploy scripts from other projects, which has the nice side effect of making them consistent again. But boy do we need to DRY these up.

If these changes look good, I'll apply them to all build and deploy scripts (at least for the projects used by krakenstein–there may be some for which they aren't as applicable).


tie-rack:
I would suggest against the "latest" tag. If it can just be _any build_, that makes it a moving target without semantic meaning. Since `pull` will try to grab `latest` if there's no tag specified, I'd suggest either not using the latest tag at all (so a user _must_ specify), or only tag production releases as latest (so the user gets something useful and stable).

cap10morgan:
@tie-rack I like the "make `latest` equivalent to `production`" idea. I'm going to do that.

PR #33, opened 2016-02-08
Merge addl voting-method keys on import

cap10morgan:
This allows us to bring in :acceptable-forms from the state defaults,
for example.

This should fix [this bug](https://www.pivotaltracker.com/story/show/113094283) for elections imported after this lands. We will still need to fix the already-imported elections, though.


cap10morgan:
We have now fixed the already-imported elections. What say you, @danielglauser?

dev-src/election_works/import.clj
```diff
@@ -143,7 +143,7 @@
                       (into {})))
         defaults (->hash defaults)
         overrides (->hash overrides)]
-    (-> (merge defaults overrides) vals set)))
+    (-> (merge-with merge defaults overrides) vals set)))
```
danielglauser: Looks like this will merge the top level maps and submaps one level deep. Do we need more of a recursive solution that handles any level of nesting? If the answer is no I suggest adding a comment.

danielglauser:
Other than my one comment LGTM

PR #34, opened 2016-04-19
I18n instructions

cap10morgan:
This adds support for translating instructions to English and Spanish, and it requests all translations from election-authority-works when it imports those.

For now it only sends English translations in election create and update events so that notifications don't break. Once we add i18n to notifications, we should remove the filters on those so they get all translations. Those are [here](https://github.com/democracyworks/election-works/blob/i18n-instructions/src/election_works/handlers.clj#L90-L91) and [here](https://github.com/democracyworks/election-works/blob/i18n-instructions/src/election_works/handlers.clj#L108-L109).

I tested this with i18n election-authority-works in krakenstein and got my vote-tomorrow email intact.

[Pivotal card](https://www.pivotaltracker.com/story/show/114350623)


docker-compose.yml
```diff
@@ -8,7 +8,7 @@ app:
     - wildfly
     - datomic
 wildfly:
-  image: quay.io/democracyworks/wildfly:8.2.0.Final
+  image: quay.io/democracyworks/wildfly:9.0.2.Final-debug
```
tie-rack: What's up with using the debug version?

src/election_works/entities/election.clj
```diff
@@ -240,6 +242,14 @@
                      merge)}
    [:fields]))
 
+(defn entity->instructions
+  [instructions]
+  (when (seq instructions)
+    (reduce (fn [m i]
+              (assoc-in m [(:instruction/type i) (:i18n/language i)]
+                        (:instruction/text i)))
+            {} instructions)))
```
tie-rack: The `seq` check is essentially performed by `reduce`. The body could be:

```clj
(reduce (fn [m i]
          (assoc-in m [(:instruction/type i) (:i18n/language i)]
                    (:instruction/text i)))
        nil instructions)
```

docker-compose.yml
cap10morgan: That's what we use everywhere (including production). It logs at the DEBUG level.

cap10morgan:
@tie-rack Ready for re-review.

cap10morgan:
@tie-rack ping

tie-rack:
You are clear for takeoff.

cap10morgan:
We can merge and deploy this after we have merged and deployed [this election-authority-works PR](https://github.com/democracyworks/election-authority-works/pull/29) and re-imported the election authorities w/ the new i18n format (currently in the "i18n States" tab in the Google Sheet).

It will then import voter reg. methods and voting methods from election-authority-works in both languages when importing new elections. But existing elections won't be updated. I don't _think_ we need to worry about that yet as turbovote-web (for better or worse) isn't currently looking at the elections for these strings. That's something we'll need to reconcile eventually (there's little point in allowing an override of those things in elections if we're just going to ignore the overridden values).

ANYway, I'll put together a deployment plan for all of this, because it's becoming a bit complex.

PR #35, opened 2016-04-27
[TH] enable looking up authorities/voting-methods for DC on import

tank157:
Part of fixing [this DC bug](https://www.pivotaltracker.com/story/show/117245961)


tank157:
Assigning to @cap10morgan 

We can now determine that `ocd-division/country:us/district:dc` is a `:state` level authority.

cap10morgan:
LGTM

PR #36, opened 2016-04-29
Allow for states like ND who don't have reg deadlines

danielglauser:


dev-src/election_works/import.clj
```diff
@@ -54,14 +54,19 @@
...
-                        (row->registration-deadline-type row)
-                        (row->by-mail-registration-deadline row)}
+  (let [by-mail-method (when (by-mail-registration-method? row)
+                           {:type :by-mail
+                            (row->registration-deadline-type row)
+                            (row->by-mail-registration-deadline row)})
         election-day-method (when (election-day-registration? row)
                               {:type :election-day})]
-    (set (keep identity #{by-mail-method election-day-method}))))
+    (set (keep identity [by-mail-method election-day-method]))))
```
cap10morgan: Why is this a vector now? Order is insignificant and there should be no dupes. Seems better represented as a set like it was before. I know it gets converted to a set anyway, but why change it here?

danielglauser: If both are nil then (as is the case with ND) constructing a set with the literal syntax goes boom. It should probably be a list because order doesn't matter but it's a vector because I didn't want to quote it.

cap10morgan: Ah, true! Good point. OK, vector seems fine then.

cap10morgan:
LGTM

PR #37, opened 2016-05-04
Add :i18n/language attribute

cap10morgan:
This will allow us to set all instructions to :en before deploying the new
i18n code that expects languages on them.

I tested this in a local krakenstein and it seemed OK (i.e. I added {:i18n/language :en} attrs to every election but with otherwise master code in election-works and went through a turbovote-web signup process). Signup went fine and I saw it schedule my notifications.


tank157:
Decided to go with a REPL based migration to set this on existing instructions rather than a database function based migration?

cap10morgan:
Yeah, I figured no need to put a fn into the transactor for a one-off change like this. It's pretty simple; I have the code for it on my dev machine.

tank157:
LGTM

PR #39, opened 2016-05-19
Add upcoming elections endpoint

cap10morgan:
I want to query for all upcoming elections from outside this component so that I can use that to verify and report on electorates, scheduled notifications, etc. based on those elections.

We'll be using this in the tool to check for which election notifications didn't get scheduled this week.

Part of this [Pivotal card](https://www.pivotaltracker.com/story/show/119913437)


resources/config.edn
```diff
@@ -1,15 +1,12 @@
 {:rabbitmq {:connection {:host #resource-config/env "RABBITMQ_PORT_5672_TCP_ADDR"
                          :port #resource-config/edn #resource-config/env "RABBITMQ_PORT_5672_TCP_PORT"}
             :queues {"election-works.ok" {:exclusive false :durable true :auto-delete true}
-                     "election-works.election.read" {:exclusive false :durable true :auto-delete false}
-                     "election-works.election.read.v2" {:exclusive false :durable true :auto-delete false}
```
tank157: Yay, bye-bye .v2s

tank157:
LGTM

PR #40, opened 2016-07-13
upgrade queue namespace, move config, fix channels

tank157:
Three Primary things in this commit:
1 - Rework queue to more recent standards
2 - Move the config lookup to core, which makes it
easier to interact with queue from a REPL
3 - Most important, don't broadcast election.delete
and election.update with events pulled from the
create channel.

[Pivotal Card](https://www.pivotaltracker.com/story/show/122617581)

Assigning to @cap10morgan


cap10morgan:
LGTM

PR #41, opened 2016-08-02
broadcast full election on delete

tank157:
Broadcasts the full election data on delete now, not just the ID. This could be useful in the future for "election cancellation" notifications that come after we've already told people about an election. If we just delete the data first, without sending it forward, it's hard to tell someone "The Foo Election we told you about on XX/YY/ZZZZ is no longer happening" since the title and date and all that would be gone.

Assigning to @ddellacosta


src/election_works/handlers.clj
```diff
@@ -126,12 +126,16 @@
...
   (s/validate delete-schema message)
-  (e.election/delete-election (db/connection) (db/db) id)
-  (async/put! deletes-channel id
-              (partial log-put!-result
-                       id
-                       "broadcast election.delete event"))
-  {:status :ok})
+  (if-let [election (e.election/read-election (db/db) (:id message))]
+    (do
+      (println "Found an election to delete" election)
```
ddellacosta: Do we want to log this instead of `println`?

ddellacosta:
Other than the println thing--which I'll leave to you how to handle, not sure 100% what we want there--LGTM.

src/election_works/handlers.clj
tank157: Yeah, probably, I'll fix that, or maybe just remove it all together.

PR #50, opened 2016-12-08
add rebroadcast namespace

tank157:
This is to help us rebroadcast the election create events from when the new cluster was less stable


dev-src/election_works/rebroadcast.clj
```diff
@@ -0,0 +1,55 @@
...
+
+(declare create-event-chan)
+(declare read-chan)
+(declare upcoming-chan)
+
+(defn init-rabbit-connections
+  [rabbit-config]
+  (def create-event-chan (async/chan))
+  (def read-chan (async/chan))
+  (def upcoming-chan (async/chan))
```
ddellacosta: So yeah, per Slack discussion not sure if there's a better way to do this than what you've done.  At the least this allows you to cleanly start this up within a single function, so 👍

dev-src/election_works/rebroadcast.clj
```diff
@@ -0,0 +1,55 @@
...
+                                                   create-event-chan)]
+    (fn []
+      (doseq [resource [read upcoming broadcast rabbit-conn]]
+        (when-not (rmq/closed? resource) (rmq/close resource)))
+      (doseq [resource [create-event-chan read-chan upcoming-chan]]
+        (async/close! resource)))))
+
+(defn upcoming-elections
+  []
+  (async/<!! ((wire-up/async->fn upcoming-chan) {})))
```
ddellacosta: Bit of a digression but: I almost feel like `async->fn` should include this last `async/<!!` bit itself...I guess we want to be able to have error-handling in there though...?

tank157: You might be onto something there. I think originally we had timeout handling outside of kehaar, so you had to do the `async/alt!!` malarky on your own code, and thus you needed to be taking from the channel. But now we have built in timeouts, so maybe it should get pushed inside.

PR #51, opened 2017-01-05
Upgrade to kehaar 0.8.1

tie-rack:


src/election_works/queue.clj
```diff
@@ -1,174 +1,23 @@
 (ns election-works.queue
   (:require [clojure.tools.logging :as log]
```
tank157: While we're cleaning up requires, let's remove log or use it.

PR #56, opened 2017-02-15
Election update tweaks

tank157:
So this expanded a bit beyond just updates to include creates as well. But removed use of deftx-data-fns, so yay!


tank157:
Assigning to @cap10morgan

src/election_works/entities/election.clj
```diff
@@ -516,13 +495,25 @@
...
+  an ID is sent in and not unique, throws an exception."
   [datomic-conn db params]
   (s/validate create-schema params)
   (let [id (:id params)]
     (if (and id (find-by-id db id))
       (throw (IllegalArgumentException. "Election with id already exists"))
-      (entity->election (transact-election datomic-conn params)))))
+      (let [{:keys [id db-after]} (create-election* datomic-conn params)
+            election (find-by-id db-after id)]
+        (entity->election election)))))
```
cap10morgan: Race conditions are fun. I guess it was already there, though. Are you _sure_ you don't want to turn this into a transactor fn? :)

test/election_works/handlers_test.clj
```diff
@@ -596,7 +598,7 @@
             (is (= :ok (:status results)))
             (is (= (set ddu) (get-in results [:election :after :district-divisions])))))))))
 
-(deftest update-election-test
+(deftest update-election-test-2
```
cap10morgan: Can we just consolidate these tests into one `update-election-test` deftest form? Possibly with 2 different `testing` forms wrapping them to explain each scenario?

src/election_works/entities/election.clj
tank157: Some day, yeah, but we make changes to elections so rarely that it's very far down on the list of things to deal with.

test/election_works/handlers_test.clj
tank157: How about I put a card in to deal with this? I honestly have no clue what the two different test scenarios are doing, and I need to use this to fix the broken NH election, so I don't really have a ton of extra time for things like that. I'd even rather just delete the first one since we hadn't been running it in over a year. So, card or delete?

cap10morgan:
@tank157 What do you think of my commit? I did a quick diff of the 2 and it seemed like the `-2` was a subset of `-1`. So I updated a few outdated names and kept the `-1` version.

PR #58, opened 2017-04-10
Election authority website

tank157:
[Pivotal Card](https://www.pivotaltracker.com/story/show/142520327)

Note: I just went with `website` as the name. It's attached to an election, I think that's pretty clear what it is.


tank157:
Assigning to @danielglauser for review.

PR #61, opened 2017-07-25
Use datomic.client for database reads

tie-rack:
For discussion only. **DO NOT MERGE!**

`datomic.api` is still used for writes and migrations.


tie-rack:
Not real :)

PR #62, opened 2017-10-30
adds polling-place-url and polling-place-url-shortened from state authority

ddellacosta:
This adds top-level `:election/polling-place-url` and `:election/polling-place-url-shortened` fields to the election at the time of import, pulled from the state authority. This is used in the vote-tomorrow notifications. See the other related PRs for more detail.

https://github.com/democracyworks/election-authority-works/pull/49
https://github.com/democracyworks/election-notification-works/pull/116
https://github.com/democracyworks/notification-templates/pull/54


resources/schemas/20171030104733-add-polling-place-url.edn
```diff
@@ -1,13 +1,13 @@
 [{:db/id #db/id[:db.part/db]
   :db/ident :election/polling-place-url
-  :db/valueType :db.type/uri
```
zonotope: has this already been applied to production? will changes that were edited in place like this take effect when this is deployed to production?

project.clj
```diff
@@ -3,29 +3,25 @@
   :url "https://github.com/democracyworks/election-works"
   :license {:name "Eclipse Public License"
             :url "http://www.eclipse.org/legal/epl-v10.html"}
-  :dependencies [[org.clojure/clojure "1.8.0"]
+  :dependencies [[org.clojure/clojure "1.9.0-beta4"]
                  [org.clojure/tools.logging "0.3.1"]
                  [turbovote.resource-config "0.2.1"]
                  [com.novemberain/langohr "3.7.0"]
```
cap10morgan: With the update to kehaar 0.11.x we should drop the explicit dependency on langohr. Kehaar will now bring in langohr 4.x itself.

resources/schemas/20171030104733-add-polling-place-url.edn
ddellacosta: Nope, won't be until it's merged and deployed. Shouldn't affect anything while still in the PR stage.

Oh and to be clear, this is a new schema addition/migration--I edited this in place, but only within the context of this PR. It's new as of this PR though.

ddellacosta:
@cap10morgan @zonotope I just realized that both of you commented on [my last commit in this PR](https://github.com/democracyworks/election-works/pull/62/commits/9742f6ea4f32abac592e493e0f68f85ad2f5d601) vs. the [full diff](https://github.com/democracyworks/election-works/pull/62/files) --probably why it seems like I was updating the schema migration in place Ben (when in actuality it is new). Sorry, I'm guessing that's because of how I linked to it in the `election-authority-works` PR. In any case, make sure you check out the full diff, that should make some things more clear.

zonotope:
ahhhh, that makes sense. i'll take a look at the full diff now.

cap10morgan:
Huh, weird. I saw all of the changes and left a review of them. No idea why it decided to attach it to that commit.

PR #64, opened 2018-01-16
Make maybe-assoc handle numbers too

cap10morgan:
Import got broken by parsing the population field into ints b/c `maybe-assoc` was assuming everything was a string. Now it doesn't assume that.


ddellacosta:
Whoops, that's my mistake--sorry about that.

PR #71, opened 2018-08-08
Add early voting

li-at-dw:
https://www.pivotaltracker.com/n/projects/930636/stories/159441387

allow :early-voting as a new type of :voting-method


src/election_works/entities/election.clj
```diff
@@ -485,6 +485,11 @@
     (s/optional-key :end         ) s/Inst
     (s/optional-key :note        ) s/Str
     (s/optional-key :instructions) instructions-schema}
+   #(and (= :early-voting (:type %)))
```
theJohnnyBrown: Don't need the `and` here, I think.

theJohnnyBrown:
Mostly looks good, just a minor nitpick

test/election_works/entities/election_test.clj
```diff
@@ -107,7 +107,20 @@
...
+                    {:type :early-voting
+                     :primary false
+                     :excuse-required false}))))
+
+  (testing "but trying to set early-voting as a primary option DOES NOT validate"
+    (is (thrown? Throwable
+                 (s/validate voting-method-schema
+                             {:type :early-voting
+                              :primary true
+                              :excuse-required false}))))
```
theJohnnyBrown: Yay tests

PR #72, opened 2018-08-11
add deadline for online voter registration method

theJohnnyBrown:
[pivotal card](https://www.pivotaltracker.com/story/show/159441015)

Adds the deadline to register online to the election import and election reads, in a manner consistent with the handling of other deadlines.


src/election_works/entities/election.clj
```diff
@@ -149,6 +153,7 @@
       :voter-registration-method.by-mail/deadline.received
       :voter-registration-method.by-mail/deadline.postmarked
       :voter-registration-method.by-mail/deadline.in-person
+      :voter-registration-method.online/deadline.online
```
tank157: it seems a bit odd to have the type essentially be `online/deadline.online`, but I suppose it mirrors the other deadlines where we include some qualifiers. Seems like online is (oddly enough) less asynchronous that `by-mail` and we could drop that last `.online` but totally optional suggestion.

theJohnnyBrown: ¯\_(ツ)_/¯ thanks though, good point

PR #73, opened 2018-08-14
Early voting dates

li-at-dw:
https://www.pivotaltracker.com/n/projects/930636/stories/159441066


dev-src/election_works/import.clj
```diff
@@ -45,6 +45,28 @@
          :ballot-request-deadline-received
          (clojure.instant/read-instant-date ballot-request-deadline)}})))
 
+(defn assoc-filled-instant
+  [map key val]
+  (if (str/blank? val)
+    map
+    (assoc map key (clojure.instant/read-instant-date val))))
```
mrichards42: Looks like you and jonathan wrote [similar functions](https://github.com/democracyworks/election-works/pull/72/files#diff-b67f7568642f9f0f847b465aa61b2d1dR54). It might make sense to have `row-get-instant` return nil for blank strings, and use `maybe-assoc` to handle adding to the map, instead of this single-purpose function.

dev-src/election_works/import.clj
```diff
@@ -34,7 +34,7 @@
   [row]
   (= "TRUE" (row-get row "excuse-required")))
 
-(defn row->voting-methods
+(defn row->ballot-request-deadline
```
mrichards42: I'd call this `row->by-mail-voting-method` since it's returning the whole voting-method map, not just the ballot-request-deadline field

dev-src/election_works/import.clj
```diff
@@ -45,6 +45,28 @@
...
+            (assoc-filled-instant :early-voting-end (row-get row "early-voting-end"))
+            (#(if (empty? %1) %1 (assoc %1 :type :early-voting :primary false))))]
+    (if (empty? early-voting)
+      #{}
+      #{early-voting})))
+
+(defn row->voting-methods
+  [row]
+  (clojure.set/union (row->ballot-request-deadline row)
+                     (row->early-voting-method row)))
```
mrichards42: Both of the `row->xx-voting-method` functions return a set with 0 or 1 maps, but I think they would be simpler and clearer if they returned either `nil` or the map. Then you would filter out the nil values and build the set in this function, similar to how `row->voter-registration-methods` works.

dev-src/election_works/import.clj
mrichards42: Continuing with this idea, the whole ballot-request function above could be simplified to:

```clj
(when-let [ballot-request-deadline (row-get-instant row "deadline:ballot-request")]
  {:type :by-mail
   :primary false
   :excuse-required (excuse-required? row)
   :ballot-request-deadline-received ballot-request-deadline})
```

resources/schemas/20180814000000-add-early-voting.edn
```diff
@@ -0,0 +1,3 @@
+[{:db/id :voting-option/type
+  :db/doc "The type of voting-option (:in-person, :by-mail, :early-voting)."
+  :db/ident :voting-method/type}]
```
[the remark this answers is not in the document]
li-at-dw: yes. this is just updating the doc

dev-src/election_works/import.clj
```diff
@@ -34,28 +48,34 @@
...
+     :ballot-request-deadline-received ballot-request-deadline}))
+
+(defn row->early-voting-method
+  [row]
+  (when-let [early-voting
+             (-> {}
+                 (maybe-assoc :early-voting-start (row-get-instant row "early-voting-start"))
+                 (maybe-assoc :early-voting-end (row-get-instant row "early-voting-end"))
+                 (#(if (empty? %1) %1 (assoc %1 :type :early-voting :primary false))))]
+    early-voting))
```
tank157: So it's ok that the result of this could be an empty hash?

li-at-dw: Oh, but I see what you mean.  I should take out the case where it could return {}.  I guess it's not wrong, strictly speaking, but nil would be better.

dev-src/election_works/import.clj
```diff
@@ -34,28 +48,34 @@
...
+     :ballot-request-deadline-received ballot-request-deadline}))
+
+(defn row->early-voting-method
+  [row]
+  (when-let [early-voting
+             (-> {}
+                 (maybe-assoc :early-voting-start (row-get-instant row "early-voting-start"))
+                 (maybe-assoc :early-voting-end (row-get-instant row "early-voting-end"))
+                 (#(if-not (empty? %1) (assoc %1 :type :early-voting :primary false))))]
+    early-voting))
```
mrichards42: In general I think having an anonymous function in a threading macro makes the code harder to follow. I'd either  pull that out into a named function, or do the `if` after the `let`, like

```clj
(let [early-voting (-> {} ...)]
  (when (seq early-voting)
    (assoc early-voting :type early-voting :primary false)))
```

I read [this article](https://stuartsierra.com/2018/07/06/threading-with-style) on threading macro style recently, and it was helpful to me. He's pretty opinionated about how they should be used, so take it with a grain of salt, but I mostly agree with his points.

li-at-dw:
Troy, Mike, and I slacked over the issue of the data schema, and came to the conclusion of going with Mike's suggestion.  However, when I went through the code and our current election-import spreadsheets again, I realized that most likely :voting-method/in-person.start and :voting-method/in-person:end are not currently in use.  (The import spreadsheet doesn't even contain the corresponding fields.)  Given that's the case, it seems that we don't necessarily know that start and end are general characteristics common to all possible voting-methods.   

Since we don't feel a lot of confidence about changing the schema to be more general, I decided that, for now, we will stay with :early-voting/start and :early-voting/end.  We can revisit this design at a future time, probably with the help of others who are more familiar with the original intention, as well as know about other voting-methods and associated attributes which we might want to add in the future.

dev-src/election_works/import.clj
```diff
@@ -18,6 +18,20 @@
 
 (def row-get (comp str/trim #(get %1 %2 "")))
 
+(defn row-get-instant [row column-name]
+    (-> row
+      (row-get column-name)
+      (#(if (str/blank? %1) nil (clojure.instant/read-instant-date %1)))))
```
mrichards42: I mean to also point this out as another anonymous function in a threading macro that could be factored out into a `let` and a `when`

dev-src/election_works/import.clj
```diff
@@ -34,28 +48,34 @@
...
+     :ballot-request-deadline-received ballot-request-deadline}))
+
+(defn row->early-voting-method
+  [row]
+  (when-let [early-voting-dates
+             (-> {}
+                 (maybe-assoc :early-voting-start (row-get-instant row "early-voting-start"))
+                 (maybe-assoc :early-voting-end (row-get-instant row "early-voting-end")))]
+    (if-not (empty? early-voting-dates)
+      (assoc early-voting-dates :type :early-voting :primary false))))
```
mrichards42: 👍 this reads more clearly to me. A couple of things that would make this more idiomatic if you want: `when-let` can be `let` here, since `early-voting-dates` will always be truthy; and `if-not` can be `when-not` since there isn't an else clause.

src/election_works/entities/election.clj
```diff
@@ -307,6 +318,12 @@
     :ballot-return-deadline-in-person
     (:voting-method.by-mail/ballot-return-deadline.in-person entity)
 
+    :early-voting-start
+    (:voting-method.early-voting/start entity)
+
+    :early-voting-end
+    (:voting-method.early-voting/end entity)
```
mrichards42: I'm OK with having these namespaced in the db, but I'd rather these be just `:start` and `:end` in our code, since all of the other keys are un-namespaced here (e.g. `:voting-method.by-mail/ballot-return-deadline.received` is just `:ballot-return-deadline-received` in the resulting map). If you have a map with `{:type :early-voting}` already, you know that `:start` means `:early-voting/start` in the db.

PR #74, opened 2018-08-17
fail fast on election import

theJohnnyBrown:


cap10morgan:
So you want to fail the rest of the import if one throws an exception?

theJohnnyBrown:
Yes. There are cases where an error will happen, but electorate-works is still producing electorates and the election create event has been sent, so if we proceed in this case we run the risk of overloading the system. This is especially bad if the error occurs on multiple elections.

cap10morgan:
OK, but why does an exception make that problem worse? Does it go faster for some reason when there is an error?

PR #80, opened 2018-12-14
Validate imports

mrichards42:
A little bit of clean up (we haven't been using the the `results` atom), and some changes so that elections are validated before they are imported, so that validation errors are clearer.

Two reasons for this change, based on the most recent election import:

1. I had both validation errors, and errors thrown b/c 0 electorates were created. This makes it easier to differentiate between the two.
2. Since errors stop the election import, it's nice to be able to root out potential errors ahead of time, so you can let the election import run unattended. This code validates all the elections before it starts to import any of them, so you can fix that class of errors ahead of time.


theJohnnyBrown:
👍 LGTM

PR #81, opened 2019-01-17
In person reg deadline

li-at-dw:
https://www.pivotaltracker.com/n/projects/930636/stories/163134044


mrichards42:
Do we need to disallow `:instructions` here and in election-authorities?

li-at-dw:
heh.  I kind of erred on the side of keeping things the way they are. I know product doesn't want to support it right now, but it's an optional field, and it seems to make sense to keep it similar to the other registration methods we have.

PR #82, opened 2019-02-13
All mail elections

li-at-dw:
https://www.pivotaltracker.com/n/projects/930636/stories/159937349

This is to allow us to specify that the primary-voting-method of an election is to be `:by-mail`, regardless of the default voting-methods specified in the state level.  When the column `primary-voting-method-type` is set to `:by-mail`, the import script modifies the voting-methods of the state, making sure that `by-mail` is the indeed the primary voting method.

Also, some side work to make it easier to write tests for the import script.


dev-test/election_works/import_test.clj
```diff
@@ -130,4 +113,57 @@
...
+                :primary true,
+                :type :by-mail}
+               {:excuse-required false,
+                :instructions {:voting-id "Fake CO voting ID instructions"},
+                :primary false,
+                :type :in-person}
+               {:excuse-required false, :primary false, :type :early-voting}}
+             overridden-vm))))
+  (testing "voting method override for :early-voting doesn't change voting methods"
+    (let [overridden-vm (set-primary-voting-method :early-voting fake-voting-methods)]
```
mrichards42: Should it be an error to specify anything other than `:by-mail` since we don't do anything with it? We could validate the primary-voting-method field in the import csv.

dev-src/election_works/import.clj
```diff
@@ -66,6 +66,10 @@
   (set (remove nil? [(row->by-mail-voting-method row)
                      (row->early-voting-method row)])))
 
+(defn row->primary-voting-method
+  [row]
+  (edn/read-string (row-get row "primary-voting-method-type")))
```
mrichards42: maybe just `"primary-voting-method"` since you don't use `type` anywhere else in the code

dev-src/election_works/import.clj
```diff
@@ -263,22 +285,25 @@
...
-                              row district-ocd-ids state-authority)}
-        (maybe-assoc :google-election-id (row-get row "google-election-id"))
-        (maybe-assoc :population (population row))
-        (maybe-assoc :website (row-get row "website"))
-        (maybe-assoc :polling-place-url (:polling-place-url state-authority))
-        (maybe-assoc :polling-place-url-shortened
-                     (:polling-place-url-shortened state-authority)))))
+  ([row]
+   (row->election row memoized-find-authority memoized-find-default-voting-methods))
+  ([row authority-fn voting-method-fn]
```
mrichards42: Most of our code that's written like this has the dependency fns listed first, which allows you to do things like:
```clj
(def my-row->election (partial my-authority-fn my-voting-method-fn))
(my-row->election row)
```

An [example](https://github.com/democracyworks/election-works/blob/532d4658bd3d977d5b7230617fc558388620124d/src/election_works/handlers.clj#L146-L162)

dev-src/election_works/import.clj
```diff
@@ -221,8 +225,25 @@
...
         authority-level-key)))
 
+(defn set-primary-voting-method
+  "If the primary-voting-method is specified, override the
+  settings currently in voting-methods.  Depending on the primary-voting-method,
+  this could entail overriding other properties in voting-method.
+  "
+  [primary-voting-method voting-methods]
+  (case primary-voting-method
+    :by-mail
```
mrichards42: I can't remember if requiring a `:by-mail` voting method is part of the election/election-authority spec, but it might be worth being defensive here:
```clj
(if (some #(= :by-mail (:type %)) voting-methods)
  ;; make :by-mail the primary method
  ;; otherwise throw an exception
  )
```

dev-src/election_works/import.clj
```diff
@@ -221,8 +225,25 @@
...
+    (set (map (fn [voting-method] (if (= (:type voting-method) primary-voting-method)
+                                    (-> (assoc voting-method :primary true)
+                                        (assoc :excuse-required false)
+                                        (dissoc :acceptable-forms)
+                                        (dissoc :ballot-request-deadline-received))
```
mrichards42: I just read through [this schema](https://github.com/democracyworks/election-works/blob/532d4658bd3d977d5b7230617fc558388620124d/src/election_works/entities/election.clj#L519-L538) and since I don't really know schema you probably know better than I do -- if a key is absent from a map schema, is it allowed to be anything, or is it not allowed in the map?

I think I'm reading that if `:by-mail` is `:primary` you can't have those two keys, but it might help to have a small comment to make that clear.

dev-test/election_works/import_test.clj
```diff
@@ -32,49 +32,32 @@
...
-                      "early-voting-start" "2015-11-01"
-                      "early-voting-end" "2015-11-02"
-                      }
-          result (row->election parsed-row)]
-      (assert (not= ::timeout (async/<!! authority-search-result)))
-      (assert (not= ::timeout (async/<!! default-voting-method-result)))
+          authority-fn (constantly [state-authority])
+          voting-method-fn (constantly fake-voting-methods)
+          parsed-row multi-district-parsed-row
+          result (row->election parsed-row authority-fn voting-method-fn)]
```
mrichards42: ❤️ this is a lot clearer than it used to be

dev-src/election_works/import.clj
li-at-dw: yeah, if a key is absent from the schema, it isn't allowed in the map.  If we want to allow it, we have to make it optional--but I checked with Esha, and she said that if we have by-mail as the primary voting method, those other fields wouldn't make any sense.

dev-src/election_works/import.clj
li-at-dw: ahh...that's a good catch...

eggsyntax:
Thanks for the work on tests btw!

dev-src/election_works/import.clj
mrichards42: 👍 that makes sense to me. can you stick a brief comment that explains why we need to dissoc those keys?

mrichards42: oh, sorry I didn't notice that you had already added info about that to the docstring

PR #87, opened 2019-03-26
No dup authority info

li-at-dw:
This PR reworks the election import and read process so that the stored election data object does not duplicate properties which are part of the election authority.  Instead, only those properties which override the election authority properties are stored in the db.  To be specific:
1.  At import, we no longer merge state authority properties (voting-methods and voter-registration-methods) into the election data object.
2.  At read, we _will_ merge the state or local authority properties into the election data object.

The motivations for this PR are
1.  In order to support by-mail local authorities and by-mail elections, we have to support the logic where the `:primary-voting-method` remains unknown until we know the address of the electorate.  It is therefore necessary to be clear about the source of information for `:primary-voting-method`, and delaying the merge of election-authority into elections until read time makes this process cleaner.

2. Since we always merge in the latest information on the authority at read time, we make sure that the election data object doesn't get stale because an authority has been updated.  This in turn would propagate to electorates and election-notification-works, and should reduce the need to re-import elections in order to pick up changes to the election authorities.


src/election_works/entities/election.clj
```diff
@@ -572,10 +613,14 @@
...
    {:type                                 (s/eq :call-clerk)}))
 
+(def primary-voting-method-source-schema
+  (s/enum :state :local :election))
+
 (def election-district-division-schema
   {:ocd-id division-regex
    :election-authority-level authority-levels-schema
    :voter-registration-authority-level authority-levels-schema
+   (s/optional-key :primary-voting-method-source) primary-voting-method-source-schema
```
mrichards42: I think we're going to want to distinguish between the schema for datomic elections and the schema for elections that are returned from `election-works.election.read` (and search). Some of the newly optional keys above are optional for datomic, but definitely required for the election data model -- we're actually having some similar conversations over in the turbovote-specs PR.

This is tricky, and I'm not sure yet what the best path forward is for this PR, but I think we want to be able to validate the full election data structure after merging. It might be as simple as validating an election + state authority before we import it, or as complicated as validating an election + state authority _and_ validating the election + each local authority before we import. And we probably also want to validate that any election authorities we import will still produce valid elections for each of the elections we have in the api currently.

This also brings up a problem w/ past elections using new authority data, which could produce incorrect results. I wonder if/how we might think about _that_, which sounds pretty ugly. Datomic at least lets us consider this as a possibility -- something like query for authorities w/ a db [`as-of`](https://docs.datomic.com/on-prem/clojure/index.html#datomic.api/as-of) the date of the election? Still sounds like some pitfalls

mrichards42: If we wanted to keep the same schemas, we could think about something like this (but honestly this feels pretty hacky):
```clj
(defn voting-method-schema* [sometimes-optional-fn]
  (s/conditional
   #(= :in-person (:type %))
   {:type                          (s/eq :in-person)
    :primary                       s/Bool
    :excuse-required               (s/eq false)
    (sometimes-optional-fn :excuse-required) (s/eq false)
    (s/optional-key :start       ) s/Inst
    (s/optional-key :end         ) s/Inst
    (s/optional-key :note        ) s/Str
    (s/optional-key :instructions) instructions-schema}
   ...))

(def datomic-voting-method-schema (voting-method-schema* s/optional-key))
(def model-voting-method-schema (voting-method-schema* s/required-key))
```

li-at-dw: Right.   The question of how to specify the schemas pre-merge and post-merge was a big one for me.  I started writing them out and then just stopped because I had too many questions.

I also realized that the current schema is lacking in some specification--for example, can we have multiple `voting-methods` designated as `primary`?  What if none of them are `primary`?  Currently we also allow duplicate types in the set of `voting-methods`, and surely that's not intentional (maybe?).

li-at-dw:
Closing this PR in favor of https://github.com/democracyworks/election-works/pull/90

PR #90, opened 2019-04-14
election overrides

li-at-dw:
For this ticket https://www.pivotaltracker.com/n/projects/930636/stories/159937349

Election data structure changes:
* Add a section of the data structure for storing voting-methods by local election authority.  This allows us to support local election authorities with voting-methods which differ from the state authority.

The election data structure stands like so (the new keys are `local-voting-methods` and `primary-voting-method-source`)
```
{
  :description "CA Election", 
  :id #uuid "5cb35046-e299-4b6e-864c-5f70fe0f980c"
  :date #inst "2019-05-26T00:00:00.000-00:00", 
  :polling-place-url "https://voterstatus.sos.ca.gov", 
  :polling-place-url-shortened "https://tvote.org/2m2Vi4e", 
  :district-divisions 
  #{{:ocd-id "ocd-division/country:us/state:ca", 
     :voter-registration-authority-level :county, 
     :election-authority-level :county, 
     :voting-methods #{...voting-methods according to state authority}
     :voter-registration-methods #{...voter-registration-methods}}, 
     :primary-voting-method-source :state}}
  :local-voting-methods #{{:ocd-id "ocd-division/country:us/state:ca/county:alameda"
                           :voting-methods #{...voting-methods for voters with elec auth ocd-id}
                           :primary-voting-method-source :local}
}
```

The following columns are added to the import spreadsheet:
* vm-override-authorities:  a comma separated list of ocd-ids of local election authorities with voting-method overrides
* supported-voting-methods: designation of an election override of the supported voting methods.  This is a comma separated list of voting-method-types like `:in-person`, `:by-mail`
* primary-voting-method: designation of an election override of the primary voting method.

(All of the above are optional.)


resources/schemas/20190403-add-election-overrides.edn
```diff
@@ -0,0 +1,93 @@
...
+
+ {:db/id #db/id[:db.part/db]
+  :db/ident :election/import-property
+  :db/valueType :db.type/ref
+  :db/cardinality :db.cardinality/many
+  :db/doc "properties of just the election"
+  :db/isComponent true
+  :db.install/_attribute :db.part/db}
+
+ ]
```
mrichards42: I like that we're storing what are essentially election-specific overrides somewhere so we can add authority updates later. 👍 

I'm a little wary of coupling this data model too tightly to the current election import column names. What about using a more abstract concept like "election overrides" or something like that?

resources/schemas/20190403-add-election-overrides.edn
```diff
@@ -0,0 +1,93 @@
...
+
+ {:db/id #db/id[:db.part/db]
+  :db/ident :import-property/supported-voting-method
+  :db/valueType :db.type/keyword
+  :db/cardinality :db.cardinality/many
+  :db/doc "Voting methods which are supported in this election (:by-mail, :in-person, :early-voting)"
+  :db.install/_attribute :db.part/db}
+
+ {:db/id #db/id[:db.part/db]
+  :db/ident :import-property/vm-local-auth-ocd
```
mrichards42: Does this need to be spelled out explicitly here, or could we use the set of `:alternative-voting-method/ocd-id`s?

resources/schemas/20190403-add-election-overrides.edn
```diff
@@ -0,0 +1,93 @@
+[{:db/id #db/id[:db.part/db]
+  :db/ident :import-property/early-voting-start
```
mrichards42: I'm wondering, rather than create additional attributes for import, if it would be possible/make sense to build up a partially filled voting methods/reg methods data structure using the existing attributes. Like:

```clj
{;; ... existing election data ...
 :override {;; most of these fit into the existing data model
            :voter-registration-methods #{{:type :online
                                           :deadline-online #inst ""}
                                          {:type :by-mail
                                           :deadline-received #inst ""
                                           :deadline-postmarked #inst ""
                                           :deadline-in-person #inst ""}}
            :voting-methods #{{:type :by-mail
                               :ballot-request-deadline-received #inst ""
                               :excuse-required false
                               :primary true}
                              {:type :early-voting
                               :start #inst ""
                               :end #inst ""}}
            ;; and then a couple need special handling
            :allowed-voting-methods #{:in-person :by-mail}
            :allowed-registration-methods #{:in-person :by-mail :online}}}
```

resources/schemas/20190409-add-election-auth-voting-methods.edn
```diff
@@ -0,0 +1,25 @@
+[
+ {:db/id #db/id[:db.part/db]
+  :db/ident :election/alternative-voting-method
```
mrichards42: I think it would be good to leave the door open to storing more than just voting-methods in this data structure. What about `:election/local-election-authority` or something like that? Maybe the lower-level keys are `:local-authority/ocd-id`, etc., or honestly we could probably just re-use the existing `:election-district-division/ocd-id`, `:election-district-division/voting-method`, etc)

resources/schemas/20190403-add-election-overrides.edn
li-at-dw: Yeah, I’m ok with a different name.  I think I started with “election overrides”, and then decided they weren’t really overrides, per se, but just extra properties that the election has.

mrichards42: Yeah, I can understand why "overrides" isn't eactly right either -- "election template"?

resources/schemas/20190409-add-election-auth-voting-methods.edn
li-at-dw: So, I started with that too, where I essentially copied the voting-methods from the authority to this data structure.  And then I decided to merge the election properties onto the local election authority voting-methods under this tree, since, by doing that, 
1.  We would validate everything about the voting methods for every authority of interest at import time;
2.  We don’t need to merge voting-methods at read time.
3.  The consumer of the election data can look up which voting-methods to use fairly easily.

Since it wasn’t clear to me what would happen with registration methods, and whether overrides with registration methods would co-occur with the voting-methods, I decided to punt how to do that..

resources/schemas/20190403-add-election-overrides.edn
li-at-dw: I left it this way because it’s possible for the import to list an authority, but the authority doesn’t have voting-methods.  In which case, I would leave it out of `:alternative-voting-method`, but not here, under the general principle of wanting to know what it is that the election import specified.

resources/schemas/20190403-add-election-overrides.edn
li-at-dw: The old code had that structure of creating voting methods out of the deadlines, and then merging them with the authority voting methods.  I moved away from it because I think the election should not be able to create voting methods that didn’t exist under the election authority. I was writing tests, and kept creating corner cases where the voting-methods set got expanded after a merge, or voting-methods disappeared unintentionally.

resources/schemas/20190403-add-election-overrides.edn
mrichards42: Should we allow listing an authority in the import if it doesn't have voting-methods, or should that be an error?

resources/schemas/20190403-add-election-overrides.edn
mrichards42: > I think the election should not be able to create voting methods that didn’t exist under the election authority.

Absolutely, you shouldn't even be able to import an election if it creates invalid voting methods. I've had elections fail validation b/c they specified `early-voting-start` in an authority that doesn't have early voting, for instance.

resources/schemas/20190403-add-election-overrides.edn
li-at-dw: I battled on ☝️ for a bit, and am still waffling.  What if we expect an authority to have an override, but it changed it's mind at some point?  I see that that import list as a list of authority updates to listen to (in addition to the state).

li-at-dw: OTOH, it could be that whoever listed that OCD really expected local authority overrides, and would appreciate an error.

resources/schemas/20190403-add-election-overrides.edn
li-at-dw: I guess I'm ok with template, property, attribute, etc.  My guess is you are objecting to using "import" as part of the keyword.  I think I tried it with `election-property` at some point, but then it kind of begged the question of why not push all of the properties to the top level, along with description, date, etc.   But then I wanted them to be sequester in their own trees, and then only thing that they actually have in common are that they were part of the import.

resources/schemas/20190409-add-election-auth-voting-methods.edn
mrichards42: I agree that it's a good plan to fully merge election properties into this data structure, I was talking about naming only.

But I realize now that election authorities and registration authorities can be at different levels, and we really _do_ need to treat these as different concepts (e.g. CT uses county-level reg authorities and the state as its election authority)

resources/schemas/20190403-add-election-overrides.edn
mrichards42: > My guess is you are objecting to using "import" as part of the keyword.

Partly, but it's also that there's some repetition in there with things that we already have a data model for, and I'm concerned that creating a very similar data model will add to maintenance.

E.g. `:import-property/reg-deadline-type` and `:import-property/reg-by-mail-deadline` are concepts that we have invented to make it possible to use a flat data structure for storing election data, but those two actually map to either `:voter-registration-method.by-mail/deadline.postmarked` or `:voter-registration-method.by-mail/deadline.received`. So now we need additional specs for `reg-by-mail-deadline` and `reg-deadline-type`, instead of being able to use the existing specs for `deadline.postmarked` and `deadline.received`.

li-at-dw: Well, that's a good point, since everything under this tree is, strictly speaking, redundant somehow....

resources/schemas/20190403-add-election-overrides.edn
mrichards42: That's a good thought too. Maybe product wants to make that call?

eggsyntax:
Just got Li to walk me through it so that I can do more coherent review, and I'll just put a summary here of what I suggested:

To me it seems like what's being stored in `:import-properties` above is essentially an _event_, and rather than stuffing it into the election (and changing it if we're given new info by authorities), we should record it as a separate import-event (/election-change-event/whatever) (presumably in datomic, but could in theory also be a separate event log). The series of relevant events can then be the single source of truth about this election, and the `election` entity is just a materialized view that can be recreated by replaying those events in order. We could connect these events either by having the election store an ordered list of event-ids, or by having each event include an election-id and a timestamp. 

The motivation here from my perspective is: a) not having multiple (potentially conflicting) sources of truth of equal weight; b) distinguishing conceptually between elections and chunks of info about elections; c) keeping the election data structure clean; d) avoiding possible confusion; e) having a clear domain-aware audit trail (ie avoiding the various issues with using datomic history for domain-sensitive audit). 

Up to you @li-at-dw whether that seems like a direction worth going, just seemed worth throwing out there.

eggsyntax:
Separate thought, re
> The following columns are added to the import spreadsheet:

> -  vm-local-auth-ocds: a comma separated list of ocd-ids of local election authorities with voting-method overrides
> -  supported-voting-methods: designation of an election override of the supported voting methods. This is a comma separated list of voting-method-types like :in-person, :by-mail

What if multiple local authorities want to override, but in different ways (eg different primary voting method)? Can that happen, and if so, is there a way to represent it on the spreadsheet?

mrichards42:
> To me it seems like what's being stored in :import-properties above is essentially an event, and rather than stuffing it into the election (and changing it if we're given new info by authorities), we should record it as a separate import-event (/election-change-event/whatever) (presumably in datomic, but could in theory also be a separate event log). The series of relevant events can then be the single source of truth about this election, and the election entity is just a materialized view that can be recreated by replaying those events in order. We could connect these events either by having the election store an ordered list of event-ids, or by having each event include an election-id and a timestamp.

@eggsyntax from my perspective that's pretty close to what's being proposed in the PR, except that there can only be one "election-change-event" at a time (so not really event sourcing). `:import-properties` is immutable, and if you need to change an election, you would create a new `:import-properties` entity, and swap it in for the old one on the election.

I'm not entirely convinced we need full-blown event sourcing here -- I could certainly be convinced, but I think we would want to tackle that as a separate project. I think we can get away with only being aware of the most recent "create/import event", since it should contain all the election-specific information we need to know. And then the election itself is still a materialized view of an election import + authority data.


> The motivation here from my perspective is: a) not having multiple (potentially conflicting) sources of truth of equal weight; b) distinguishing conceptually between elections and chunks of info about elections;

Right now `:import-properties` only includes data about voting- and reg-methods, but doesn't include things like a description, date, district-divisions, etc. and I buy that it could make sense to have a data structure that includes everything that is in the import. I was starting to suggest thinking of `:import-properties` as more of an "election template" earlier in this review, and this makes me wonder if we really do want this to be something like a template -- i.e. it's the same election data structure, but sparsely populated, and then the materialized election is created by merging it w/ the authority data.

> c) keeping the election data structure clean;

You have to link the materialized election to the import event somehow, and `:import-properties` (or whatever we end up calling it) is just a ref, so it's still its own conceptual entity. But I can see that "election" is still the top-level entity, and "import-properties", while its own thing, "belongs" to the election. Is that what makes this seem not clean?

mrichards42:
> What if multiple local authorities want to override, but in different ways (eg different primary voting method)? Can that happen, and if so, is there a way to represent it on the spreadsheet?

Local authorities can specify whatever voting methods they want (they use the same data structure as the state authority). The primary-voting-method and supported-voting-methods fields are for election-level overrides

eggsyntax:
> > What if multiple local authorities want to override, but in different ways (eg different primary voting method)? Can that happen, and if so, is there a way to represent it on the spreadsheet?
> 
> Local authorities can specify whatever voting methods they want (they use the same data structure as the state authority). The primary-voting-method and supported-voting-methods fields are for election-level overrides

Oh, right, ok; I was conflating their intention with `vm-local-auth-ocds`. NM!

eggsyntax:
> I'm not entirely convinced we need full-blown event sourcing here -- I could certainly be convinced, but I think we would want to tackle that as a separate project. I think we can get away with only being aware of the most recent "create/import event", since it should contain all the election-specific information we need to know. And then the election itself is still a materialized view of an election import + authority data.
> 
OK, that's fair, I was imagining that successive import events might be sparser, and would be merged or deep-merged onto the existing election. But if each one is comprehensive, there's no need for that except to have a domain-level audit trail, which certainly doesn't seem to be as high-priority.

> Right now `:import-properties` only includes data about voting- and reg-methods, but doesn't include things like a description, date, district-divisions, etc. and I buy that it could make sense to have a data structure that includes everything that is in the import. 

It does seem like if we want to think of them as events we'd want to include everything.

> I was starting to suggest thinking of `:import-properties` as more of an "election template" earlier in this review, and this makes me wonder if we really do want this to be something like a template -- i.e. it's the same election data structure, but sparsely populated, and then the materialized election is created by merging it w/ the authority data.
>

That seems like a reasonable direction to me too --  encourage product to just fill it the fields that have changed, and generate a sparse event from that.

> > c) keeping the election data structure clean;
> 
> You have to link the materialized election to the import event somehow, and `:import-properties` (or whatever we end up calling it) is just a ref, so it's still its own conceptual entity. But I can see that "election" is still the top-level entity, and "import-properties", while its own thing, "belongs" to the election. Is that what makes this seem not clean?

Ummm no, that's pretty much a brain-fart on my part, you're totally right that it's a separate entity already. Does seem like we'd want to make it a non-component in datomic, so that `pull` wouldn't automatically populate it, and so that it wouldn't be auto-deleted if the election were deleted.

All in all I'm happy with whichever direction we go (as-is, event-log-of-full-template, sparse-event-log). For me the main mental flash was just _thinking_ of this as an event with its own existence, not as something that's conceptually part of the election. I'm ok with that translating into actual changes or not, now or later :)

resources/schemas/20190403-add-election-overrides.edn
```diff
@@ -0,0 +1,93 @@
...
+  :db/cardinality :db.cardinality/many
+  :db/doc "ocd-ids of local authorities with voting-methods which can override state voting-methods."
+  :db.install/_attribute :db.part/db}
+
+ {:db/id #db/id[:db.part/db]
+  :db/ident :election/import-property
+  :db/valueType :db.type/ref
+  :db/cardinality :db.cardinality/many
+  :db/doc "properties of just the election"
+  :db/isComponent true
```
eggsyntax: See comment on main page -- I think it might be worth making `isComponent` false, so that a `pull` won't autopopulate it, and so that (to the extent we want to preserve them as event-log-ish things expressing the intent of the election authority) they won't be auto-deleted if the election is deleted.

resources/schemas/20190403-add-election-overrides.edn
li-at-dw: Product wants to have an error, so that's what we'll do!

resources/schemas/20190403-add-election-overrides.edn
li-at-dw: OK. After thinking about this for awhile....While I see that it is an extra set of complexity for these properties to be listed separately in the data model, I also think that it is not obvious (and possibly not desirable) to assume that every field in the election import has to map to a specific place in the election data structure.   I think it's a good principle to not couple the import data fields to the rest of the actual election data structure too tightly.  

My preference is to leave the import properties under their own tree for now (with a different name, perhaps--like `template-properties` or `static-parameters` or something more conceptual), and make the rest of the data structure specify the election fully without referencing this part.  I think we can make the `read-election` handler return the election object without `import-property`.   

So, for example, at this moment, I haven't added `primary-voting-method-source` anywhere, figuring that it can be computed based on `import-property` by any downstream source.  But, it's probably better to have that logic sit inside `election-works` rather than in the consumers of the data structure.

mrichards42: What would you think of adding the `:alternative-voting-methods` part of the data structure in this PR, but handling the `:import-properties` in a separate PR? I think the first part is pretty straightforward, but I suspect based on the discussion so far that we have some thinking to do about how best to represent the static/template/import data.

I'm definitely on board with it being its own entity, but I have concerns with replicating the import spreadsheet's data structure in datomic, since I think the spreadsheet is not a great format.

> So, for example, at this moment, I haven't added primary-voting-method-source anywhere, figuring that it can be computed based on import-property by any downstream source. But, it's probably better to have that logic sit inside election-works rather than in the consumers of the data structure.

I think fields that can be derived from import-properties and the authority data should be represented only in the materialized election -- I think you're saying that we should store that piece of data on the election itself, rather than asking downstream consumers to figured it out, and I agree.

li-at-dw: > What would you think of adding the `:alternative-voting-methods` part of the data structure in this PR, but handling the `:import-properties` in a separate PR? I think the first part is pretty straightforward, but I suspect based on the discussion so far that we have some thinking to do about how best to represent the static/template/import data.

I think that's a good idea.  We can lump the `:import-properties` type of work as part of a ticket to update the election structure according to authority changes.

> I think fields that can be derived from import-properties and the authority data should be represented only in the materialized election -- I think you're saying that we should store that piece of data on the election itself, rather than asking downstream consumers to figured it out, and I agree.

Yes, I think we agree on this.

dev-src/election_works/import.clj
```diff
@@ -296,25 +170,73 @@
 
 (def validate-election (partial s/validate e.election/create-schema))
 
+(defn row->import-property [row]
```
mrichards42: This feels like a great starting point for an import refactor, along the lines of how `election-authority-works` has an [intermediate](https://github.com/democracyworks/election-authority-works/blob/master/src/election_authority_works/import/intermediate.clj) representation.

I'm wondering if it might make sense to call this function `parse-row` and have it take in a `csv-row` and return an intermediate representation, e.g.

```clj
(defn parse-row [csv-row]
  (-> {:description (row-get row "description")
       :date (row-get-instant row "date")
       :ocd-ids (str/split (row-get csv-row "district-divisions") #"\s+")}
      (maybe-assoc ,,,)
      (maybe-assoc ,,,)
      ,,,))
```

And then we can think of concepts like `csv-row` (unparsed), `row` (parsed), and `election` (final data structure).

resources/schemas/20190422-add-primary-voting-method-source.edn
```diff
@@ -0,0 +1,8 @@
+[
+ {:db/id #db/id[:db.part/db]
+  :db/ident :election-district-division/primary-voting-method-source
```
mrichards42: Do we need `primary-voting-method-source` on each set of voting methods, or could there be a single one on the election? I'm thinking something like `:election` or `:authority`, and the in the read handler we update it to be `:state` or `:local` based on which set of voting-methods is in use. If this one is `:state` we're already going to have to overwrite it in read when we substitute in the local voting methods, so I'm wondering if we need the alternate-voting-methods one.

Not that I'm opposed to it, and it might be that it's just easier to store the source on each set of voting methods.

src/election_works/entities/election.clj
```diff
@@ -247,17 +268,26 @@
...
-    (when (contains? params :district-divisions)
-      (for [district-division district-divisions]
-        [:db.fn/retractEntity (:db/id district-division)]))))
+    (cond-> []
+      (contains? params :district-divisions)
+      (into (for [district-division district-divisions]
+              [:db.fn/retractEntity (:db/id district-division)]))
+      (contains? params :alternative-voting-methods)
+      (into (for [alt-vms alt-voting-methods]
+              [:db.fn/retractEntity (:db/id alt-vms)])))))
```
mrichards42: I believe datomic should handle this, since `:election/alternative-voting-method` has `:db/isComponent true`, and datomic [retracts sub-components](https://docs.datomic.com/on-prem/schema.html#component) when it retracts an entity.

In fact, I'm not sure we need this function at all, since I assume `:election/district-division` is a component too.

li-at-dw: OK, I was very confused by this as well, but I think this function, even though it's called `->retractions`, is called only for `updates`.  And the thing it does is to retract the old components (during a update), without retracting the top level entity.

mrichards42: oooh, ok in that case this makes sense, ignore me.

resources/schemas/20190422-add-primary-voting-method-source.edn
li-at-dw: um.  Yeah, I guess with the way our logic works, what you're saying would work as well. I'm happy to change that.

But, I was hoping that we don't have to change the `read` method at all.  So, if `election-notification-works` is reading the election, it should be figure out which set of voting-methods to use based on the user's local election-authority, and go from there.  I am weary of manipulating the objects inside `:district-divisions`, because there could be multiple of those...

src/election_works/voting_methods.clj
```diff
@@ -0,0 +1,81 @@
...
+                 :ballot-request-deadline-received ballot-request-deadline)
+    :early-voting
+    (-> voting-method
+        (maybe-assoc :start early-voting-start)
+        (maybe-assoc :end early-voting-end))
+    voting-method))
+
+(defn merge-voting-methods
+  [defaults election-props]
+  (map (partial assoc-props election-props) defaults))
```
mrichards42: To me it feels like this part of the code relies on details of the import spreadsheet that I don't think we want to creep into the main codebase.

A couple of thoughts:

1. I think we want to throw an error if the import includes data for unsupported voting-methods (e.g. if if has early-voting-start or end when there is no early-voting method).

    1a. actually this brings up a question about deadlines -- it seems like different jurisdictions might have different deadlines; (a) I might be wrong about this, and (b) I suppose that could be an exceptional case where we need to think of those as different elections. But the data model would support different deadlines by jurisdiction, so it would be nice if we could figure out a reasonable way to do that at some point in the future, perhaps when we tackle a richer election import interface.

2. I still think it might be more flexible (and possibly cleaner) to build partially filled voting methods from the election import, and then `merge` them into the authority voting methods (similar to the previous `merge-voting-methods` function). We could still `throw` if a voting method shows up in the import that doesn't exist in the authority.

I remember you talking about issues with accidentally removing or adding voting methods with that function. If we wanted to do this, I would suggest something like this:

```clj
(-> (filter-allowed-voting-methods base-voting-methods #{:by-mail :in-person :whatever})
    ;; throw if there are methods that are not in base-voting-methods
    (merge-voting-methods election-deadline-vms)
    (select-primary-voting-method :by-mail))
```

which is almost exactly what you have now, but with the filtering done before the merging, so you can tell if the election-specific vms include bad data and throw if they try to expand voting-methods.

resources/schemas/20190422-add-primary-voting-method-source.edn
mrichards42: oh, I thought the plan was to change the read handler so the consumers didn't have to do extra work? or was the idea that we would do the manipulation in `electorate-works`? I guess it might help caching if we were to do it in electorate-works.

Your call, I don't think I have a strong opinion on this until I see how it's going to be used in notys or in tv-web.

li-at-dw: I think it makes more sense if it were to happen in `electorate-works`.  In some ways, if we were to do it in `election-works`, we end up duplicate some logic in `electorate-works` to figure out the local election authority, and so on.

src/election_works/voting_methods.clj
```diff
@@ -0,0 +1,81 @@
...
+  [primary-voting-method voting-methods]
+  (case primary-voting-method
+    :by-mail
+    (if (some #(= :by-mail (:type %)) voting-methods)
+      (map #(if (= (:type %) :by-mail)
+              (assoc % :primary true)
+              (assoc % :primary false))
+           voting-methods)
+      (throw (Exception. (str ":by-mail is not a valid voting-method and cannot be made as the primary voting-method." ))))
+    voting-methods))
```
mrichards42: I'm wondering if it makes sense to have the logic be the same no matter the `primary-voting-method`, i.e. set that one as `:primary true` and all the other ones as `:primary false`, and throw if the method you're trying to set as primary doesn't exist.

I'm not sure we have a use-case right now for setting anything but `:by-mail` as the primary voting method, but if for instance, an election tried to set `:in-person` as the primary voting method, right now we're ignoring that.

mrichards42:
A couple thoughts on naming, which are definitely suggestions and not mandates, since all naming is subjective:

1. I think something like `:local-authority-voting-methods` would be more precise than `:alternative-voting-methods`
2. Wherever we're using `auth` to mean `authority` I think we should go ahead and spell it out. "auth" means "authentication" or "authorization" to me, and I it tripped me up a couple times. If the variables are getting too long I think I'd rather see `vms` to mean `voting-methods` rather than `auth` to mean `authorities`.

src/election_works/voting_methods.clj
li-at-dw: yeah, I guess I just wasn't ready to generic set  the primary voting method to just anything yet.  Like, if we set `:by-mail` to be primary, we have to go back and make sure `excuse-required` is false, and get rid of the acceptable forms.  And I didn't know if we were ready to consider what it would mean, for example, if someone set `:early-voting` to be primary.

mrichards42: That motivation makes sense. I'm fine with this function handling primary-voting-methods differently, but it should throw if you give it a non-nil primary-voting-method that we don't know how to handle.

I believe the schema also considers something a `:primary` `:early-voting` method as invalid, so we could also handle it that (and if it doesn't, we should fix that). I'd say as much as possible the import code should handle whatever input you give it, but throw if it produces an invalid election -- and that could happen explicitly here, or later when we validate the whole election against the schema.

mrichards42: also fwiw unless the exception type you're throwing has some kind of specific meaning, I'd say it's a little more idiomatic to do:
```clj
(throw (ex-info "some message" {:some :relevant, :extra data}))
```

src/election_works/voting_methods.clj
li-at-dw: I went down the road of doing this, but now I remember why I changed the whole merge in the first place.  Your state authority voting methods and election authority voting methods don't have to contain the same voting-method types.  So, maybe the state-authority supports early-voting, in-person, and by-mail, but the local authority is all-mail.   So you specify early-voting dates for the state authority, but the local authority only supports by-mail, and so when you do this merge, you are stuck.

mrichards42: Has product talked about this sort of situation? Initially it feels like relaxing the import spec is a bad idea . . . would it work to throw an exception if the election supplies data that _none_ of the authority voting methods use? I've definitely had imports where the spreadsheet included data for vms that the state authority doesn't handle, and that threw an error -- but maybe the research team is actually OK with us ignoring that bad data?

If we needed to restrict the election vms to only the ones the authority supports, we could do something like this:
```clj
(let [authority-hash (->hash authority-vms)
      election-hash (select-keys (->hash election-vms) (keys authority-hash))])
```

li-at-dw: OK, I talked to Esha, and we agreed that we can throw an exception if _none_ of the authority voting methods used a piece of data.

src/election_works/entities/election.clj
```diff
@@ -201,15 +198,34 @@
...
+           primary-voting-method-source]
+    :as local-voting-method}]
+  (when local-voting-method
+    (tx/dissoc-optional-nil-values
+     {:local-voting-method/ocd-id ocd-id
+      :local-voting-method/primary-voting-method-source
+      primary-voting-method-source
+      :local-voting-method/voting-method
+      (map voting-method->tx-map voting-methods)}
+     [])))
```
mrichards42: looks like there aren't any `optional-nil-values` here to dissoc?

src/election_works/entities/election.clj
```diff
@@ -224,17 +240,22 @@
...
       :election/polling-place-url-shortened
-      polling-place-url-shortened}
+      polling-place-url-shortened
+      :election/local-voting-method
+      (map local-voting-method->tx-map local-voting-methods)}
+
      [:election/google-election-id :election/population :election/website
-      :election/polling-place-url :election/polling-place-url-shortened])))
+      :election/polling-place-url :election/polling-place-url-shortened
+      :election/local-voting-methods])))
```
mrichards42: `method` not `methods`?

src/election_works/entities/election.clj
```diff
@@ -427,8 +457,29 @@
...
              (map entity->voter-registration-method)
-             set)}
-   [:voting-methods :voter-registration-methods]))
+             set)
+
+    :primary-voting-method-source
+    (or (:election-district-division/primary-voting-method-source entity)
+        :state)}
+   [:voting-methods :voter-registration-methods
+    :primary-voting-method-source]))
```
mrichards42: we're guaranteeing that `primary-voting-method-source` isn't going to be nil above, right?

src/election_works/entities/election.clj
```diff
@@ -427,8 +457,29 @@
...
+
+    :voting-methods
+    (some->> entity
+             :local-voting-method/voting-method
+             (map entity->voting-method)
+             set)
+
+    :primary-voting-method-source
+    (:local-voting-method/primary-voting-method-source entity)}
+   [:ocd-id :voting-methods :primary-voting-method-source]))
```
mrichards42: not 100% sure how this gets used, but I think all 3 of those are required if we have a local-voting-method entity

src/election_works/entities/election.clj
```diff
@@ -7,20 +7,15 @@
             [clj-time.core :as time]
             [clj-time.coerce :as time.coerce]
             [com.climate.newrelic.trace :refer [defn-traced]]
-            [clojure.tools.logging :as log])
+            [clojure.tools.logging :as log]
+            [election-works.entities.election-schema :refer :all])
```
mrichards42: if it isn't too much hassle, I'd prefer that we don't `:refer :all` and instead make a namespace alias here (like `election-schema` or even just `schema`

src/election_works/voting_methods.clj
```diff
@@ -0,0 +1,83 @@
...
+  (:require [clojure.tools.logging :as log]
+            [schema.core :as s]
+            [election-works.utils :refer [maybe-assoc]]))
+
+(defn merge-voting-methods
+  "Merges election-vms onto authority-vms, based on :type and :primary.
+  Keep only those election-vms which are
+  already in authority-vms."
+  [authority-vms election-vms]
+  (let [->compound-key-map #(select-keys % [:type :primary])
```
mrichards42: I know this was in the original, but I think it's a mistake to include `:primary` as one of the keys. I don't know this code as well as you, so I might be wrong, but I'm thinking of a situation where the election overrides the primary voting method, so you end up trying to merge:

```clj
;; election
#{{:type :by-mail :primary false} {:type :in-person :primary true}}
;; authority
#{{:type :by-mail :primary true} {:type :in-person :primary false}}
;; result
#{{:type :by-mail :primary false} {:type :in-person :primary true} {:type :by-mail :primary true} {:type :in-person :primary false}}
```

src/election_works/voting_methods.clj
```diff
@@ -0,0 +1,83 @@
...
+  [voting-methods primary-voting-method]
+  (s/validate primary-voting-method-schema primary-voting-method)
+  (case primary-voting-method
+    :by-mail
+    (if (some #(= :by-mail (:type %)) voting-methods)
+      (-> (map #(if (= (:type %) :by-mail)
+                  (assoc % :primary true)
+                  (assoc % :primary false))
+               voting-methods)
+          schema-compliant-all-mail-vm)
```
mrichards42: I find this and `schema-compliant-all-mail-vm` a little hard to read partially b/c of the long anonymous functions, but also b/c we're mapping over the full voting method set twice.

what about something like
```clj
;; instead of mapping over everything in schema-compliant-all-mail-vm
;; this is just the logic for making the :by-mail type the primary method
(defn ->all-mail [vm]
  (-> vm
      (assoc :primary true)
      (assoc :excuse-required false)
      (dissoc :acceptable-forms)
      (dissoc :ballot-request-deadline-received))

;; and then in select-primary-voting-method
(for [vm voting-methods]
  (if (= :by-mail (:type vm))
    (->all-mail vm)
    (assoc vm :primary false)))

File: src/election_works/voting_methods.clj [hunk 18]
```diff
@@ -0,0 +1,83 @@
...
+  (if (:primary-voting-method election-props)
+    :election
+    default-authority-level))
+
+(defn ->voting-methods
+  "Create the voting-methods for the election by combining
+  the voting-methods of the relevant state or election authorities
+  and merging them with the election-specific properties"
+  [authority-vms election-vms {:keys [supported-voting-methods primary-voting-method]}]
+  (some-> (keep-supported-voting-methods supported-voting-methods authority-vms)
```
mrichards42: if an election specifies voting method data that isn't in `supported-voting-methods` do we catch that anywhere?

I'm thinking of a case where an election only supports, e.g. `in-person`, but it then includes some `by-mail` deadlines. since we're only filtering authority vms, the election vms we're merging in won't be filtered.

src/election_works/entities/election.clj
```diff
@@ -302,17 +304,31 @@
...
-        [:db.fn/retractEntity (:db/id district-division)]))))
+    (cond-> []
+      (contains? params :district-divisions)
+      (into (for [district-division district-divisions]
+              [:db.fn/retractEntity (:db/id district-division)]))
+      (and (contains? params :import-property) (:db/id import-property))
+      (into [[:db.fn/retractEntity (:db/id import-property)]])
+      (contains? params :alternative-voting-methods)
+      (into (for [alt-vms alt-voting-methods]
+              [:db.fn/retractEntity (:db/id alt-vms)])))))
```
eggsyntax: I'd love some kind of separation between test / consequence forms. It's really easy for readers to get confused by that lack of distinction. There's not a perfect option, but I would consider:
- putting them on the same line even thought the line might be really long
- forcing indentation on the consequence, at the risk that it'll get overridden by other people's editors
- putting blank lines between pairs
- It's unconventional, but this is maybe my favorite, given that commas are whitespace, and editors won't screw it up:
```clojure
(cond-> []
  (contains? params :district-divisions)
  , (into (for [district-division district-divisions]
            [:db.fn/retractEntity (:db/id district-division)]))
  (and (contains? params :import-property) (:db/id import-property))
  , (into [[:db.fn/retractEntity (:db/id import-property)]])
  ...)
```

mrichards42: I'd definitely come down on the side of the clojure style guide that [suggests](https://github.com/bbatsov/clojure-style-guide#shor-forms-in-cond) blank lines or comments

src/election_works/voting_methods.clj
li-at-dw: It's in the import script.
https://github.com/democracyworks/election-works/blob/e03e43c9dd9a5e6e38463de0270c879428d36182/dev-src/election_works/import.clj#L193-L210

src/election_works/voting_methods.clj
li-at-dw: So, first of all, we change the primary voting-method on the election level, _after_ merging the election to the authority.  So, for the cases that I worked though, using `[:type :primary]` seems ok.  Whatever errors it creates will get caught on schema validation.

But secondly, I think that it's quite difficult to reason through all the corner cases of the voting-methods merge.  I changed the merge key (to be without `:primary`) at one point and then changed it back because I am not comfortable changing the merge key without knowing a lot more about the combinations of the election authority and elections and what the intention was. Some of that would definitely require more conversations with product and so on.

P.S. The original code also had `excuse-required` as a key and I did drop that after verifying with Esha that we don't need an `excuse-required` column in the election import.

test/election_works/voting_methods_test.clj
```diff
@@ -0,0 +1,135 @@
...
+                          :ballot-request-deadline-received #inst "2015-10-23T00:00:00.000-00:00"}}
+          election-props {}
+          merged-vms (->voting-methods authority-vms election-vms election-props)]
+      (is (=  #{{:type :early-voting, :primary false, :excuse-required false
+                 :end #inst "2015-11-02T00:00:00.000-00:00",
+                 :start #inst "2015-11-01T00:00:00.000-00:00"}
+                {:type :by-mail, :primary true, :excuse-required false,
+                 :instructions {:voting-id {:en "Fake CO by-mail voting ID instructions"}}}
+                {:type :in-person, :primary true, :excuse-required false,
+                 :instructions {:voting-id {:en "Fake CO voting ID instructions"}}}}
```
mrichards42: looks like we ended up with both `:by-mail` and `:in-person` as `:primary` voting methods. I also see that `:ballot-request-deadline-received` isn't present in merged-vms, but that might be b/c is was removed in `->all-mail`?

test/election_works/voting_methods_test.clj
```diff
@@ -0,0 +1,135 @@
...
+                          :ballot-request-deadline-received #inst "2015-10-23T00:00:00.000-00:00"}}
+          election-props {:primary-voting-method :by-mail}
+          merged-vms (->voting-methods authority-vms election-vms election-props)]
+      (is (=  #{{:type :early-voting, :primary false, :excuse-required false
+                 :end #inst "2015-11-02T00:00:00.000-00:00",
+                 :start #inst "2015-11-01T00:00:00.000-00:00"}
+                {:type :by-mail, :primary true, :excuse-required false}
+                {:type :in-person, :primary false, :excuse-required false}}
+              merged-vms))
+      (s/validate #{voting-method-schema} merged-vms))))
```
mrichards42: I think we want to test a couple other scenarios:

1. all-mail state authority where the election includes by-mail deadlines (should fail validation I think?)
3. supported-voting-methods that excludes the primary voting method (should fail validation)

mrichards42: (unless those cases are covered in the import tests and I didn't see them)

test/election_works/voting_methods_test.clj
li-at-dw: oh, that's a bad test. the authority-sms had two primary methods.  Let me fix that.

li-at-dw: `:ballot-request-deadline-received` is not merged during the merge process because that field corresponds to voting-method `{:type by-mail :primary false}`.   (And we have logic in the merge which only merges in election keys (`:type` and `:primary`) which already exist in the authority.)

PR #92, opened 2019-07-01
Add early voting url to election-works import script

dziring:
https://app.asana.com/0/1127682668253243/1127683047512328/f

Adds the ability to have a URL associated with an election that has early-voting. Paired with @saramartinez  on this change.

Please comment with any feedback or things we may have missed!


src/election_works/entities/election.clj
```diff
@@ -336,6 +340,10 @@
           (= :early-voting (:voting-method/type entity))
           (:voting-method.early-voting/end entity))
 
+    :url
+    (if (= :early-voting (:voting-method/type entity))
+        (:voting-method.early-voting/url entity))
```
mrichards42: Minor bit of style: usually (at least at DW) we use `when` instead of `if`, unless there's an else branch. It isn't a huge deal, but for larger blocks of code it can be helpful to see a when and know that it means there isn't an else.

mrichards42: Also a (very) minor thing -- indentation is off here.

dev-src/election_works/import.clj
```diff
@@ -199,14 +209,15 @@
...
+                      [:early-voting-url :early-voting :url]
                       [:ballot-request-deadline :by-mail :ballot-request-deadline-received]]
         unused-info (for [[row-key vm-type vm-key] check-params
                           :let [vm-with-type (filter #(= (:type %) vm-type) all-vms)]
                           :when (and (row row-key)
                                      (= (count (filter #(% vm-key) vm-with-type)) 0))]
                       row-key)]
     (if (> (count unused-info) 0)
-      (throw (ex-info "Voting-method dates were provided but unused"
+      (throw (ex-info "Voting-method dates or url were provided but unused"
```
mrichards42: It didn't occur to me when we were looking at the code earlier that this piece already handles the case where you have an early-voting-url without a start or end date. Why don't we let this code handle that error, and ignore it in `->early-voting-deadlines`?

It might be clearer to change this error message too, in case we want to check more than voting methods. Maybe something like `(str "Unused election-import columns: " (str/join ", " unused-info))`

dziring: My only concern with relying on this as the error checking is that it means it can't easily be tested, since the other code happens prior to the datomic transaction, and this happens after. Is there any integration testing that could be leveraged? What are your thoughts?

mrichards42: This happens before the transaction. It's called in [`read-elections`](https://github.com/democracyworks/election-works/blob/d45204b0541ad4fcf02ec3a801959652214093e1/dev-src/election_works/import.clj#L311).

That said, this function is actually about making sure that the voting-methods in the election spreadsheet will be used by one of the authorities (state or local), not that the data in the spreadsheet is consistent with itself, so . . . I don't mind checking in both places if you want to leave it alone.

dziring: Ah, looking closer, if I'm understanding correctly (please correct me if not), it's the `find-authority` code that reads authorities from datomic that is hanging within the `read-elections` function. Regardless, for now I'll leave this part as is

mrichards42: Right, usually we would injected external services as a fn, like:
```clj
(defn find-authority [ocd-id] ...)

(defn read-elections
  ;; the default case
  ([filename] (read-elections find-authority filename))
  ([authority-fn filename]
    ;; the actual code
    ))
```

So that in a test you can mock that service easily:
```clj
(let [authority-fn (constantly {:authority :data})]
  (read-elections authority-fn filename))
```

The code in this ns is kind of messy and doesn't do that everywhere (one of the drawbacks of dependency-injection is that you have to keep injecting those dependencies everywhere you use them, which can get tedious).

dev-src/election_works/import.clj
```diff
@@ -136,12 +136,14 @@
...
-      (when-not (and (:start early-voting-dates) (:end early-voting-dates))
-        (throw (Exception. (str "Early voting url given without both early-voting start and end
-                                 dates. Start: " start " End: " end " Url: " url)))))
 
-    (when-not (empty? early-voting-dates)
+    (when (and url (not (and start end))
+      (throw (Exception. (str "Early voting url given without both early-voting start and end dates."
+                              " Start: " (or start "nil")
+                              " End: " (or end "nil")
+                              " Url: " url)))))
```
mrichards42: This is a perfect case for `ex-info`, since the exact start/end/url aren't necessarily important to have in the error message:
```clj
(ex-info "Early voting url given without both early-voting start and end dates."
         {:start start :end end :url url})
```

dev-src/election_works/import.clj
```diff
@@ -217,7 +217,8 @@
                                      (= (count (filter #(% vm-key) vm-with-type)) 0))]
                       row-key)]
     (if (> (count unused-info) 0)
-      (throw (ex-info "Unused election-import columns." {:unused-columns unused-info})))))
+      (throw (ex-info (str "Unused election-import columns: " (str/join ", " unused-info))
+                      {:unused-columns unused-info})))))
```
dziring: @mrichards42 If I'm understanding right, this would include this info in the string, but also provide in a structured way if needed -- is this the proper usage?

mrichards42: Yes, that's right exactly. In this case it's helpful to have the unused columns listed out, but also nice to have structured data.

dev-src/election_works/import.clj
```diff
@@ -131,18 +131,28 @@
...
-                 (maybe-assoc :end end))]
-    (when-not (empty? early-voting-dates)
-      (assoc early-voting-dates :type :early-voting :primary false))))
+  [start end url]
+  (let [early-voting-dates
+        (-> {}
+            (maybe-assoc :start start)
+            (maybe-assoc :end end))]
+
+    (when (and url (not (and start end))
```
mrichards42: based on the fact that those are two `maybe-assoc`s above, should this be `(or start end)`?

dziring: The possible valid options should be:
- none (start, end, url) present
- start + end present without URL
- start + end present with URL

So based on that, I think this is correct

dziring: This is a change in behaviour from before (which would allow start or end without the other), but confirmed with Esha that this is what is expected

mrichards42: gotcha -- if that's the case, let's rewrite this function to make that clear:
```clj
(if (and start end)
  (-> {:start start :end end :type :early-voting :primary false} (maybe-assoc :url url))
  (when url (throw ...)))
```

dev-src/election_works/import.clj
```diff
@@ -131,18 +131,24 @@
...
 (defn ->early-voting-deadlines
-  [start end]
-  (when-let [early-voting-dates
-             (-> {}
-                 (maybe-assoc :start start)
-                 (maybe-assoc :end end))]
-    (when-not (empty? early-voting-dates)
-      (assoc early-voting-dates :type :early-voting :primary false))))
+  [start end url]
+    (if (and start end)
```
mrichards42: indentation

leilakass:
LGTM 👍

PR #93, opened 2019-07-11
it is ok if no electorates are created

teaforthecat:
Not all elections have users. I can see how we would like that to be true, but it isn't, yet.


dev-src/election_works/import.clj
```diff
@@ -309,7 +309,6 @@
                                                                           :election (-> elc :election (select-keys [:description :id]))))
                                   (recur (inc electorate-count)))
                                 electorate-count))]
-    (when (= electorates-created 0) (throw (Exception. (str "no electorates created for " el))))
```
dzcode: Could this just be a logged warning instead removing it entirely so there is still the sanity check?

teaforthecat: Idk, would that infer that there should be? I don't think there _should_ be. There is an info log that would say that there are zero electorates, which seems satisfactory to me.

dzcode: My 2c would be that a message like "WARNING: no electorates created. If this is a medium-large election, this may indicate a problem with the data" could help catch things, but I don't feel strongly

teaforthecat: Do you have an idea of what sort of problems with the data there could be?

dzcode: I don't specifically, and don't let this hold up the review. Maybe @eggsyntax or @saramartinez would know?

eggsyntax: I'll second Dan on that. We've certainly encountered problems in the past at times during election imports. I like the idea of not making it an exception, but I agree that a warning would be good.

teaforthecat:
closing in favor of #94

PR #94, opened 2019-07-17
Check election existence before and after import

mrichards42:
This removes the "0 electorates created" exception, and gets to the root
of the issue it was trying to solve. If something went wrong during the
import and the election was not created, the import throws an exception.

This also adds a check before importing to make sure that the election
does not already exist.

Elections are considered unique on description, date, and ocd-ids


mrichards42:
I wrote the new election-import chore doc with this in mind -- the two "import errors" in the troubleshooting section are these exceptions (and I didn't mention anything about "0 electorates" in that new doc).

teaforthecat:
This looks good. Is there a way to update elections, then?

mrichards42:
> Is there a way to update elections, then?

Good question -- not really. Depending on the type of change, we might have to reschedule notifications, so typically updating an election is deleting it (which deletes electorates and notifications) and importing another one.

For changes that don't require rescheduling notifications, we'll occasionally transact a change directly to datomic, but it's really case-by-case.

PR #96, opened 2019-09-04
Add source-type, source-notes, and source-date to elections

dzcode:
Asana card: https://app.asana.com/0/1135662569808788/1131746625558591

Housekeeping:
- [ ] Add **source-type**, **source-notes**, and **source-date** to [election spreadsheet](https://docs.google.com/spreadsheets/d/1CIhVbxb1ru-sj2lHtEFhGZfLBoQKBjhTgd2e39BB4GI/)
- [x] Let Ada know about this addition to the election spreadsheet and what is allowed in the columns


eggsyntax:
>     * [ ]  Add **source-type**, **source-notes**, and **source-date** to [election spreadsheet](https://docs.google.com/spreadsheets/d/1CIhVbxb1ru-sj2lHtEFhGZfLBoQKBjhTgd2e39BB4GI/)
> 
>     * [ ]  Let Ada know about this addition to the election spreadsheet and what is allowed in the columns

I'd probably bring it up with Ada before adding to the spreadsheet or merging the PR, just in case they have input that'll require changes.

resources/schemas/20190903000000-add-source-fields.edn
```diff
@@ -0,0 +1,19 @@
...
+  :db/valueType :db.type/string
+  :db/cardinality :db.cardinality/one
+  :db/doc "Free text notes about the source of this data"
+  :db.install/_attribute :db.part/db}
+{:db/id #db/id[:db.part/db]
+ :db/ident :election/source-date
+ :db/valueType :db.type/instant
+ :db/cardinality :db.cardinality/one
+ :db/doc "The date this election was discovered from this source"
+ :db.install/_attribute :db.part/db}]
```
mrichards42: My recollection of the discussion we had several weeks ago about this card was that we wanted to model this as a component entity, rather than top-level attributes

i.e. this:
```clj
{:election/description "My election"
 ,,,
 :election/source
 {:source/type :vip
  :source/note "hello world"
  :source/date #inst "2019-01-01"}}
```
instead of this:
```clj
{:election/description "My election"
 ,,,
 :election/source-type :vip
 :election/source-note "hello world"
 :election/source-date #inst "2019-01-01"}}
```


Among the benefits is that I believe you can model the requirements better in schema (or spec when we get there): you can say that the source entity has a required a type and date and an optional note, and an election has an optional source. And you would be able to get rid of `validate-source-fields` in the import script. We'd like to avoid additional checks in that script if possible -- ideally validations are part of the schema and/or election-create handler, and the import script uses the validations that are part of the service rather than defining its own.

src/election_works/utils.clj
```diff
@@ -7,6 +7,14 @@
 (def district-regex
   #"\Aocd\-division\/country\:[a-z]{2}\/district\:[a-z]{2}")
 
+(def valid-source-types #{:website
+                          :call
+                          :email
+                          :google-alert
+                          :vip
+                          :state-contact
+                          :state-list})
```
mrichards42: You should be able to create an enum as part of the schema

resources/schemas/20190903000000-add-source-fields.edn
```diff
@@ -0,0 +1,19 @@
+[{:db/id #db/id[:db.part/db]
+  :db/ident :election/source-type
+  :db/valueType :db.type/keyword
+  :db/cardinality :db.cardinality/one
+  :db/doc "The category for this source
+  (:website, :email, :call, :google-alert, :vip, :state-contact, :state-list)"
```
eggsyntax: Have you considered treating this as a Datomic enum? 
https://docs.datomic.com/on-prem/schema.html#enums

eggsyntax: [I hadn't seen Mike's comments when I posted mine; this comment is subsumed by his broader point]

dev-src/election_works/import.clj
```diff
@@ -198,8 +206,25 @@
         (catch NumberFormatException ex
           val)))))
 
+(defn validate-source-fields
+  [{:keys [source-type source-notes source-date] :as election}]
+  (cond (or (and source-type (nil? source-date))
+            (and (nil? source-type) source-date))
```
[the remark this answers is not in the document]
mrichards42: @eggsyntax I think you're looking for this 😉 
```clj
(defn xor [& args]
  (and (some identity args)
       (not (every? identity args))
```

resources/schemas/20190903000000-add-source-fields.edn
dzcode: That makes sense to me. Are there any other places in the code that we have required fields in datomic? Looking at the [datomic docs](https://docs.datomic.com/on-prem/schema.html#required-attributes), they seem to use `db.entity/attrs`, but I don't see that elsewhere in our code, so thinking there might be another way to do it? It seems like [predicates](https://docs.datomic.com/cloud/schema/schema-reference.html#attribute-predicates) may also be an option, but not seeing that as well

dzcode: Scratch that, I see it's done elsewhere at the schema level in code, but not in the datomic schema itself -- assuming that's what you meant?

mrichards42: Right, in the paragraph at the end I was talking about the `schema` library that we use for validation in code.

dev-src/election_works/import.clj
```diff
@@ -58,6 +60,12 @@
     (when (not (str/blank? ocds))
       (set (str/split ocds #",")))))
 
+(defn row->source-type
+  [row]
+  (let [source-type (row-get row "source-type")]
+    (when (not (str/blank? source-type))
+      (keyword source-type))))
+
```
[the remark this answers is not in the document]
mrichards42: Agreed that it wouldn't be a bad idea to have `row-get` return nil for blank strings. Though I see that `row-get` does already trim the string, so `(when (not (str/blank? field))` could just be `(when (seq field)`

dev-src/election_works/import.clj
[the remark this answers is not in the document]
eggsyntax: :joy:

dev-src/election_works/import.clj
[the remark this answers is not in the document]
dzcode: It looks like in most places the behaviour is the same in the end case if `row-get` returns nil or "", so I've changed the default to be nil and then given "" in the few cases where it matters. This is definitely much less repetitive, thanks for suggesting @eggsyntax

resources/schemas/20190903000000-add-source-fields.edn
dzcode: It looks like in the rest of the cases where e've just defined the enums at the schema layer in code and not in the database, so going to follow that pattern here

eggsyntax: Huh, I thought we did enums in some places. Unfortunately they're hard to grep for.

I wonder why we wouldn't be using them. @mrichards42 any idea?

dzcode:
There's an issue with importing the source that I'm working on, so planning a small change to the import script

resources/schemas/20190903000000-add-source-fields.edn
mrichards42: We definitely use entity enums [in](https://github.com/democracyworks/election-works/blob/master/resources/schemas/20180213001-make-authority-levels-entities.edn) [some](https://github.com/democracyworks/email-works/blob/master/resources/schemas/20170828-000-initial-schema.edn) [places](https://github.com/democracyworks/election-mail-works/blob/master/resources/schemas/20170906193830-mail-logging.edn), though not [in](https://github.com/democracyworks/election-authority-works/blob/master/resources/schemas/20151016000-default-voter-reg-methods.edn#L5-L6) [others](https://github.com/democracyworks/user-works/blob/master/resources/schemas/000-initial-schema.edn#L54-L55).

I honestly don't have a full grasp on the implications of one or the other, but the most useful difference might be whether or not we expect this to be an open or closed set. My expectation is that it's closer to open than closed. I'd also welcome feedback from @gws, @sduckett or @tank157 who might have more insight into these decisions from the past.

dev-src/election_works/import.clj
```diff
@@ -63,10 +63,14 @@
...
+(defn row->source
   [row]
-  (when-let [source-type (row-get row "source-type")]
-    (keyword source-type)))
+  (let [source (-> {}
+                   (maybe-assoc :type (when-let [source-type (row-get row "source-type")]
+                                        (keyword source-type)))
+                   (maybe-assoc :notes (row-get row "source-notes"))
+                   (maybe-assoc :date (row-get-instant row "source-date")))]
+    (when (seq source) source)))
```
mrichards42: A couple ways around `(when (seq source) source)`

1. [`not-empty`](https://github.com/clojure/clojure/blob/clojure-1.10.0/src/clj/clojure/core.clj#L5537-L5541) is literally that, and could go right at the end of the threading macro
2. you could also `assoc` onto `nil`, which might look kind of strange, but I assume `(maybe-assoc nil :k nil)` won't create an empty map . . . probably more readable to go w/ `not-empty` though

dev-src/election_works/import.clj
```diff
@@ -63,10 +63,14 @@
...
     (set (str/split ocds #","))))
 
-(defn row->source-type
+(defn row->source
   [row]
-  (when-let [source-type (row-get row "source-type")]
-    (keyword source-type)))
+  (let [source (-> {}
+                   (maybe-assoc :type (when-let [source-type (row-get row "source-type")]
+                                        (keyword source-type)))
```
mrichards42: this is a good place for `some->`, which will shortcut as soon as the value is nil

```clj
(maybe-assoc :type (some-> (row-get row "source-type") keyword))
```

dev-src/election_works/import.clj
```diff
@@ -20,43 +20,58 @@
...
 
-(def row-get (comp str/trim #(get %1 %2 "")))
+(defn row-get
+  ([row field default]
+   (let [val (-> row
+                 (get field "")
+                 str/trim)]
+     (cond
+       (not (str/blank? val)) val
+       (some? default) default)))
```
mrichards42: I think this `cond` is the same as:
```clj
(if-not (str/blank? val)
  val
  default)
```
since if nothing matches the `cond` you end up returning `nil` anyways

src/election_works/entities/election.clj
```diff
@@ -221,14 +221,24 @@
...
      ))
 
+(defn source->tx-map
+  [{:keys [type notes date] :as source}]
+  (when source
+    (tx/dissoc-optional-nil-values
+     {:source/type type
+      :source/notes notes
+      :source/date date}
+     [:source/type :source/notes :source/date])))
```
mrichards42: I thought type and date are required?

src/election_works/entities/election_schema.clj
```diff
@@ -13,6 +13,9 @@
 (def fields-schema
   {s/Keyword (s/enum :leave-blank :optional :requested :required)})
 
+(def source-type-schema
+  (s/enum :call :email :google-alert :state-contact :state-list :vip :website))
+
```
mrichards42: might want to stick this right above the `source-schema` since it's only used there.

src/election_works/entities/election.clj
dzcode: Yep, thanks for catching. I was thinking about it the wrong way, but this makes sense now

dev-src/election_works/import.clj
eggsyntax: Wow, I've _never_ noticed `not-empty`. Just when you think you're finally familiar with all the clojure core functions... :upside_down_face:

resources/schemas/20190903000000-add-source-fields.edn
sduckett: I think the interesting part of components is how retraction works: 
https://docs.datomic.com/on-prem/schema.html#component 

plus 

https://github.com/democracyworks/election-works/blob/master/resources/schemas/20180213001-make-authority-levels-entities.edn#L17-L22

eggsyntax: Enums aren't necessarily components, though, are they?

sduckett: not necessarily, but if you make them so, there are interesting behaviors to watch out for.

mrichards42: @sduckett yep, as you said, based on wes's comment there it sounds like they should _never_ be components (the enum ref itself that is).

I was mostly linking to some of those as examples of places we've chosen to use enum refs vs just keywords. Curious if other folks have thoughts on when you'd make that call.

dev-src/election_works/import.clj
```diff
@@ -219,9 +232,9 @@
                       {:unused-columns unused-info})))))
 
 (defn parse-csv-row [csv-row]
-  (-> {:description (row-get csv-row "description")
+  (-> {:description (row-get csv-row "description" "")
        :date (row-get-instant csv-row "date")
-       :district-ocd-ids (-> (row-get csv-row "district-division")
+       :district-ocd-ids (-> (row-get csv-row "district-division" "")
                              (str/split #"\s+"))}
```
mrichards42: Not important to fix, but It actually looks like this is the only place you _need_ the default `""` in `row-get`, but you could also do `(some-> (row-get csv-row "district-divison") (str/split #"\s+"))` and eliminate the need for the default empty string. From what I can tell, the only other places you're using it are on the `description` column, which will fail validation if it's nil -- which we'd want to do anyways.

dzcode: That's a good point, and actually let's me clean up the `row-get` function quite a bit by using `some->` there as well

mrichards42: 👏 that new function is super clear and simple

resources/schemas/20190903000000-add-source-fields.edn
eggsyntax: Tossed it in #datomic just to see if we get any interesting thoughts from the broader community: https://clojurians.slack.com/archives/C03RZMDSH/p1567797221099700

dzcode:
@eggsyntax When you have time, could you give my latest change a once-over before I merge -- I found it was needed while testing a bunch of imports locally 
https://github.com/democracyworks/election-works/pull/96/commits/c841bd37f004539d7d3ec9fd51f56af9384ed0a8

PR #97, opened 2019-09-11
[import] minimal response for each electorate

teaforthecat:
The electorate data can be very large and is probably occupying a lot of network bandwidth which may be causing the imports to be slower than they need to be.

The electorate that is returned is not used, so lets not return it.

This may or may not have an impact on the time it takes to import an election.  It will be hard to see if this actually has an impact unless we deploy it.


src/election_works/handlers.clj
```diff
@@ -131,7 +131,16 @@
...
+                                                    ;; - return something small
+                                                    ;; in consideration of
+                                                    ;; possible network delays -
+                                                    ;; this is used as a locking
+                                                    ;; queue so two elections
+                                                    ;; are not imported at the
+                                                    ;; same time when the
+                                                    ;; results-chan closes, the
+                                                    ;; next election is imported
+                                                    (async/>!! results-chan 1)))]
```
mrichards42: Sending less data over the wire is bound to be an improvement, though I'm suspicious that the main problem might not be the size of each message, but the total number of messages. What if we kept track of the count here with an atom and only sent back each 100th message?

teaforthecat: that is a good idea - let's try both!

mrichards42: 👍 I have some elections to import today, so I'll tell you how it goes

PR #98, opened 2019-09-17
Send electorate-create progress during import

mrichards42:
... instead of sending each electorate individually as it is created.

This is part 1 (of 2) of small changes to speed up election imports. Turns out the main issue w/ streaming back electorates to pace election imports is not the size of each individual electorate, but the total number of messages we send. I've consistently seen my laptop take about 10 messages a second from that streaming queue, while electorates are created at 20+ a second. The upshot is that election imports actually finish about twice as fast as the import script thinks they do.


src/election_works/handlers.clj
```diff
@@ -123,27 +123,28 @@
...
+                     event
+                     (fn [electorate]
+                       (swap! electorate-count inc)
+                       (when (zero? (mod @electorate-count electorate-progress-increment))
+                         (async/>!! results-chan {:electorate-count @electorate-count}))))]
     (add-watch jobs/jobs :all-electorates
                (fn [key atom old-state new-state]
                  (when-not (get new-state routing-key)
+                   (async/>!! results-chan {:electorate-count @electorate-count
+                                            :done true})
```
gws: Probably a dumb question, but where was `:done true` being sent down the pipe before?

mrichards42: It wasn't -- we were using the closed channel to indicate that the import was done before: https://github.com/democracyworks/election-works/blob/555f3b291ab08f809704a10d0804245ce4d0556b/dev-src/election_works/import.clj#L325-L331

gws: Indeed, it was a dumb question - got it, thanks!

mrichards42: Not a dumb question at all! That was a hole in the existing implementation that made it impossible to know for sure if all the electorates were actually created.

PR #100, opened 2019-10-04
Validate election-import columns

mrichards42:
We've had a misspelled column in the election spreadsheet for many months.

Even better would be to write a spec for import rows and use conform on them, but this is a good enough stop-gap for now.


mrichards42:
`election-day-registration` had a capital E

eggsyntax:
The only downside is that if they add another column, we have to remember to name it here -- anyplace we could add a line of documentation to help people remember to do so?

eggsyntax:
There's also

https://github.com/bhauman/spell-spec

mrichards42:
> The only downside is that if they add another column, we have to remember to name it here

I think that's (mostly) a feature not a bug. And I'm guessing the error will make it pretty obvious:
```
Reading election Lansing Municipal Election ...Syntax error (ExceptionInfo) compiling at (/private/var/folders/21/12kn3g916v346w2vpxwx074c0000gn/T/form-init8239648284052151728.clj:1:125).
Unexpected columns: ("Election-day-registration")
```
though extra documentation never hurts.

mrichards42:
@eggsyntax that will be super helpful once this is using spec!

PR #101, opened 2019-10-09
Date validation

solomonsokoya:
[Asana Card](https://app.asana.com/0/1144005340871649/1141683109544624/f)


teaforthecat:
I'm thinking that the schema is not the place to do this validation because we want to validate the election date, which is one schema, to other dates, which are in other schemas. Maybe we can use schema still, but the value of it is pretty low in this case.

teaforthecat:
What we could do is ship just the start and end date comparison and leave the other comparisons (which are across schemas) to another piece of work.

teaforthecat:
Let's put the date validations here: `election-works/dev-src/election_works/import.clj:213`

src/election_works/entities/election_schema.clj
```diff
@@ -20,46 +20,60 @@
 (def instructions-schema
   {s/Keyword {(s/enum :en :es) s/Str}})
 
-(def voting-method-schema
+(defn start-before-end?
+  [{:keys [start end ballot-request-deadline-received deadline-online deadline-received]}]
+  (cond
+    (and start end) (not= 1 (compare start end))
```
mrichards42: This probably works, but since the [docs](https://clojuredocs.org/clojure.core/compare) say "Returns a negative number, zero, or a positive number" it would be safer to use `(not (pos?` here.

src/election_works/entities/election_schema.clj
```diff
@@ -20,46 +20,60 @@
...
+           (s/optional-key :ballot-return-deadline-received  ) s/Inst
+           (s/optional-key :ballot-return-deadline-postmarked) s/Inst
+           (s/optional-key :ballot-return-deadline-in-person ) s/Inst
+           (s/optional-key :acceptable-forms)                  #{forms-schema}}))
 
-(def voter-registration-method-schema
+(def voting-method-schema
+  (s/constrained
+   voting-method-schema*
+   start-before-end?))
```
mrichards42: It looks like `start-before-end?` compares `start` to one of many end dates. Most of these entities don't have `start` (e.g. `:by-mail` voting-method only has `ballot-request-` and `ballot-return-` deadlines), so this test will be ignored.

How about sticking this constraint on each individual voting or reg method that it should apply to?
e.g.
```clj
;; additions marked with **
(def voting-method-schema
  (s/conditional
    #(= :early-voting (:type %))
    (s/constrained
     {:type                              (s/eq :early-voting)
      :primary                           (s/eq false)
      :excuse-required                   s/Bool
      (s/optional-key :start)            s/Inst
      (s/optional-key :end  )            s/Inst
      (s/optional-key :url  )            s/Str
      (s/optional-key :instructions)     instructions-schema}
     ;; ** start must be before end
     #(before? (:start %) (:end %)))
    #(and (= :by-mail (:type %)) (not (:primary %)))
    (s/constrained
     {:type                                               (s/eq :by-mail)
      :primary                                            (s/eq false)
      :excuse-required                                    s/Bool
      (s/optional-key :note                             ) s/Str
      (s/optional-key :instructions                     ) instructions-schema
      (s/optional-key :ballot-request-deadline-received ) s/Inst
      (s/optional-key :ballot-return-deadline-received  ) s/Inst
      (s/optional-key :ballot-return-deadline-postmarked) s/Inst
      (s/optional-key :ballot-return-deadline-in-person ) s/Inst
      (s/optional-key :acceptable-forms)                  #{forms-schema}}
     ;; ** request deadline must be before return deadlines
     #(every? (partial before? (:ballot-request-deadline-received %))
              [(:ballot-return-deadline-received %)
               (:ballot-return-deadline-postmarked %)
               (:ballot-return-deadline-in-person %)]))
     ...
     ))

```
That would make things a bit cluttered, but it would apply the constraints in the correct places.

teaforthecat:
I forgot about the create-schema* funny - that would be where to put it.

test/election_works/voting_methods_test.clj
```diff
@@ -70,6 +70,39 @@
...
+                          :end #inst "2015-11-02T00:00:00.000-00:00",
+                          :start #inst "2015-11-05T00:00:00.000-00:00"},
+                         {:type :by-mail
+                          :primary false
+                          :ballot-request-deadline-received #inst "2015-10-23T00:00:00.000-00:00"}}
+          election-props {}
+          merged-vms (->voting-methods authority-vms election-vms election-props)]
+      (is (thrown? Throwable
+                   (s/validate #{voting-method-schema} merged-vms)))))
+  (testing "->voting-methods -- to ensure end date is after start date"
```
mrichards42: This test looks like it's a duplicate of the original `->voting-methods-test`

test/election_works/voting_methods_test.clj
```diff
@@ -70,6 +70,39 @@
                  :instructions {:voting-id {:en "Fake CO voting ID instructions"}}}}
               merged-vms))
       (s/validate #{voting-method-schema} merged-vms)))
+  (testing "->voting-methods -- to ensure error will be thrown if start date is after end date"
```
mrichards42: I think this test is already covered by the ones you added in `election-schema-test`

src/election_works/entities/election_schema.clj
teaforthecat: you wouldn't get usable error messages that way, but if you added a name as a third arg to `s/constrained` like `"ballot-request-deadline-received  not before other dates"` maybe that would be good enough?

teaforthecat:
another approach would be to write a function that validates the sequential-ness the dates of the election and **then** worry about how to fit that into the schema

src/election_works/entities/election_schema.clj
solomonsokoya: I am guessing `before? = start-before-end?` ?

solomonsokoya: If we continue to go this route, would Chris suggestion be voided?

mrichards42: I think `state-before-end?` is a little too specific, since e.g. in the second example, you're really testing `ballot-request-deadline-received-before-ballot-return-deadline-received?`.

It seems like what we care about is if one date is before another, so `before?` (or maybe `on-or-before?` depending on what the constraints are) would be more accurate.

mrichards42: > If we continue to go this route, would Chris suggestion be voided?

The code in this ns only works to validate dates within a single voting- or reg-method. Looks like the card also asks you to validate that all the dates and deadlines in the election are on or before the election date itself, which we can't do like this.

So correct me if I'm wrong @teaforthecat, but at this point I think we're both suggesting a separate function that validates every voting and reg date against the election date. And then once we have that function we can figure out where it belongs. And I think at this point we're both on board w/ putting that function in the main election schema.

src/election_works/entities/election_schema.clj
```diff
@@ -20,72 +20,115 @@
...
-    (s/optional-key :note)                s/Str
-    (s/optional-key :instructions)        instructions-schema}
+   (s/constrained
+    {:type                                 (s/eq :online)
+     :url                                  s/Str
+     :supports-iframe                      s/Bool
+     (s/optional-key :deadline-online)     s/Inst
+     (s/optional-key :note)                s/Str
+     (s/optional-key :instructions)        instructions-schema}
+    #(on-or-before? (:start %) (:deadline-online %)))
```
mrichards42: It looks like `:online` reg doesn't have a `:start` key

src/election_works/entities/election_schema.clj
```diff
@@ -20,72 +20,115 @@
...
+     (s/optional-key :note)                s/Str
+     (s/optional-key :instructions)        instructions-schema
+     (s/optional-key :deadline-received)   s/Inst
+     (s/optional-key :deadline-postmarked) s/Inst
+     (s/optional-key :deadline-in-person)  s/Inst
+     (s/optional-key :acceptable-forms)    #{forms-schema}}
+    ;; ** request deadline must be before return deadlines
+    #(every? (partial on-or-before? (:deadline-received %))
+             [(:deadline-postmarked %)
+              (:deadline-in-person %)])
```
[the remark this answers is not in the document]
mrichards42: right -- same as w/ online reg, only one date should be part of any given by-mail reg method, so there's nothing to validate here (though we still want to check the date as part of the overall election validation).

src/election_works/entities/election_schema.clj
```diff
@@ -20,73 +20,100 @@
 (def instructions-schema
   {s/Keyword {(s/enum :en :es) s/Str}})
 
+(defn start-before-end?
```
mrichards42: I believe this function is unused now

src/election_works/entities/election_schema.clj
```diff
@@ -20,73 +20,100 @@
...
-    (s/optional-key :instructions) instructions-schema}
+   (s/constrained
+    {:type                          (s/eq :in-person)
+     :primary                       s/Bool
+     :excuse-required               (s/eq false)
+     (s/optional-key :start       ) s/Inst
+     (s/optional-key :end         ) s/Inst
+     (s/optional-key :note        ) s/Str
+     (s/optional-key :instructions) instructions-schema}
+    #(on-or-before? (:start %) (:end %)))
```
mrichards42: could you add the 3rd argument to `s/constrained` for each of these so we get a good error message? e.g. in this case it might be "in-person start must be on or before end"

PR #102, opened 2019-10-11
[import] validate authority

teaforthecat:
This will help to make it clear what the problem is during the import.


dev-src/election_works/import.clj
```diff
@@ -210,10 +210,10 @@
       (catch NumberFormatException ex
         val))))
 
-(defn validate-election [election row]
-  (s/validate e.election/create-schema election)
-  ;; check that the voting-method fields expressed in the spreadsheet
-  ;; were used by at least one authority.
+(defn validate-authority
```
mrichards42: How about `validate-voting-methods` since that's all it's doing now?

dev-src/election_works/import.clj
```diff
@@ -227,9 +227,14 @@
                                      (= (count (filter #(% vm-key) vm-with-type)) 0))]
                       row-key)]
     (if (> (count unused-info) 0)
-      (throw (ex-info (str "Unused election-import columns: " (string/join ", " unused-info))
+      (throw (ex-info (str "Authority disagrees with election on: "
+                           (string/join ", " unused-info)
+                           " there is a problem with the election or the authority")
```
mrichards42: This is much better error message, thanks!

dev-src/election_works/import.clj
```diff
@@ -210,10 +210,10 @@
       (catch NumberFormatException ex
         val))))
 
-(defn validate-election [election row]
-  (s/validate e.election/create-schema election)
-  ;; check that the voting-method fields expressed in the spreadsheet
-  ;; were used by at least one authority.
+(defn validate-authority
+  "check that the voting-method fields expressed in the spreadsheet
+   were used by at least one authority."
```
mrichards42: I think "used by at least one authority" might be a little misleading since we don't have a concept of "authority" in the election data structure itself once it's built -- voting methods end up under either the `:district-divisions` or `:local-voting-methods` keys. How about something like this: "Check that the voting-method fields expressed in the spreadsheet are supported by at least one authority"?

For a little context -- it would actually be a lot nicer if we could validate that all the election's voting methods are retained when the state's voting methods are merged in here: https://github.com/democracyworks/election-works/blob/ba86deceeb95a76dfbf4b41ccb895a46aef11941/src/election_works/voting_methods.clj#L6-L9

. . . but unfortunately the whole point of the work on local election authorities was so that we could support local authorities that have different voting-methods than the state defaults, so by definition they might not support all the voting methods in the election itself. This function is here to make sure that the data in the spreadsheet is used by at least _one_ of the different sets of voting-methods, since otherwise that indicates an error.

dev-src/election_works/import.clj
teaforthecat: I'm a little uncomfortable using the term voting-methods untethered to either the authority or the election because the reason for this change - it was hard to tell what was happening. If we say "voting-methods" that begs the question, which "voting-methods"? The fact that we are checking only voting methods seems insignificant. We want to see if there is any possible data entry problem on the election or the authority. Since there is no data entry for voting methods, "voting-methods" won't point to the problem as well as "authority".

dev-src/election_works/import.clj
teaforthecat: changed

dev-src/election_works/import.clj
mrichards42: 🚲 🏠 

> If we say "voting-methods" that begs the question, which "voting-methods"?

We're validating the intended voting methods (what was in the spreadsheet) against the actual voting methods (what's on the election we just built).

> We want to see if there is any possible data entry problem on the election or the authority.

I think this gets to what I find strange about calling this authority validation -- the authority is already production data, and thus already validated. We're really validating the csv that we're importing. It's true that we may need to change the authority to allow importing this election if the election is correct, but I think we have to start with the assumption that the authority is correct.

mrichards42: I don't care that much though, so pick whatever makes sense to you.

Ultimately I think we need to divide this import up a little more and create a spec for the csv since there are plenty of other errors we've been missing that could be caught with a spec.  If we had even more separation of parsing the csv and building the election, I think this sort of validation would fit squarely in the realm of csv validation, rather than election validation, since as you've pointed out, we're talking about a data entry problem, not an election that fails spec, or even a voting-method that fails spec.

PR #103, opened 2019-10-14
add check-dates! function

teaforthecat:
I believe this catches all the scenarios we are looking for, but it unfortunately raises an exception instead of letting schema handle the error message.


teaforthecat:
If this is the approach we want to take, it seems like this would be best suited for the import script so the exceptions are raised locally and not on the server.

teaforthecat:
if we put this validation function on the `create-schema` we could catch the exception and return a special error message

mrichards42:
> if we put this validation function on the create-schema we could catch the exception and return a special error message

I looked a little into schema error messages and couldn't find a way to do this, without listing each individual check as a separate constraint w/ a custom error message. Do you know what this might look like?

src/election_works/entities/election_schema.clj
```diff
@@ -133,3 +133,44 @@
...
+   [:ballot-return-deadline-postmarked :date]
+   [:ballot-return-deadline-received :date]])
+
+(defn check-dates [election]
+  (let [date (:date election)
+        dd (:district-divisions election)
+        voting-methods (map :voting-methods dd)
+        voting-registration-methods (map :voting-registration-methods dd)
+        ;; assoc the election date to make it easy to compare various dates to
+        ;; it
```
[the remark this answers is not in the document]
teaforthecat: I could add a namespace to the assoc'd keyword

teaforthecat:
I tried to build functions to pass to `s/constrained` with a name that would indicate what is going on - it got real messy real quick though because of all the nesting and the number of attributes that are dates. It is probably possible but I don't think it would be very readable. This data structure is nice and tidy.

src/election_works/entities/election_schema.clj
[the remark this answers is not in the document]
teaforthecat: done

teaforthecat:
actually, hold on, duh, schema works by raising exceptions - plus we already have `handle-errors` catching `clojure.lang.ExceptionInfo` - so all we need to do is throw an error that fits what `handle-errors` is doing

src/election_works/entities/election_schema.clj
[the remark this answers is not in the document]
mrichards42: 👍 works for me! Could we make that `:election/date` so that the error message is clearer? I think right now it would say ":end should not be after :election-works.entities.election-schema"

src/election_works/entities/election_schema.clj
```diff
@@ -133,3 +133,48 @@
   {:type source-type-schema
    :date s/Inst
    (s/optional-key :notes) s/Str})
+
+(defn check-order!
+  "Check the order of two attributes of a map. Equal is ok. Throw an exception if they are not
+  in the order they are passed in"
+  [m [a b]]
+  (when (and (m a) (m b) (pos? (compare (m a) (m b))))
+    (throw (ex-info (str "\n\nout of order exception: \n"
```
mrichards42: Any reason for the newlines?

src/election_works/entities/election_schema.clj
```diff
@@ -133,3 +133,48 @@
...
+        voting-registration-methods (map :voting-registration-methods dd)
+        ;; assoc the election date to make it easy to compare various dates to
+        ;; it, using a unique namespace
+        methods (map #(assoc % ::date date)
+                     (concat (flatten voting-methods)
+                             (flatten voting-registration-methods)))]
+
+    (doseq [method methods]
+      (doseq [[before after] dates-in-order]
+        (check-order! method [before after])))
```
mrichards42: Just a thought: would it make sense to collect all the invalid dates and then throw a single exception, in case there are multiple?

src/election_works/entities/election_schema.clj
```diff
@@ -133,3 +133,48 @@
...
+
+(defn check-order!
+  "Check the order of two attributes of a map. Equal is ok. Throw an exception if they are not
+  in the order they are passed in"
+  [m [a b]]
+  (when (and (m a) (m b) (pos? (compare (m a) (m b))))
+    (throw (ex-info (str "\n\nout of order exception: \n"
+                         a "(" (m a) ")"
+                         " should not be after "
+                         b "(" (m b) ")" "\n")
```
mrichards42: Not a blocker, but it would be nice if we could figure out which map the date came from, e.g. "in-person registration deadline should not be after election date", rather than ":end should not be after :election/date". Especially for keys that are common between multiple types, like `:start` and `:end`.

src/election_works/entities/election_schema.clj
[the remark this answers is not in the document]
teaforthecat: done

src/election_works/entities/election_schema.clj
teaforthecat: I'm used to seeing the schema errors with the important info buried deep, so this will help to highlight the problem I think.

src/election_works/entities/election_schema.clj
teaforthecat: Let's wait to see if it is a pain

teaforthecat:
Yes, I think this is the only pr needed (on accident). I still need to run it locally to make sure it works, then I'll go ahead and merge it.

mrichards42:
@teaforthecat would you be willing to restrict this PR to just doing the validations against election/date (i.e. just your three commits)? #101 is basically done, and there's been a fair amount of work in solomon's branch since you branched off it, plus this branch doesn't cover all the cases that the other PR does. I think we could merge each set of commits to master separately anyways since they're solving different problems and don't have conflicts with each other.

src/election_works/entities/election.clj
```diff
@@ -577,17 +577,19 @@
...
+    :date s/Inst
+    :district-divisions #{election-schema/election-district-division-schema}
+    (s/optional-key :google-election-id) s/Str
+    (s/optional-key :population) (s/pred integer?)
+    (s/optional-key :website) s/Str
+    (s/optional-key :polling-place-url) s/Str
+    (s/optional-key :polling-place-url-shortened) s/Str
+    (s/optional-key :local-voting-methods) #{election-schema/local-voting-method-schema}
+    (s/optional-key :source) election-schema/source-schema}
+   election-schema/check-order!))
```
mrichards42: I think your error message gets swallowed by schema if it's added as a constraint -- we might need to check it separately.

src/election_works/entities/election.clj
```diff
@@ -610,17 +612,19 @@
...
+            (s/optional-key :description) s/Str
+            (s/optional-key :date) s/Inst
+            (s/optional-key :district-divisions) (s/maybe #{election-schema/election-district-division-schema})
+            (s/optional-key :population) (s/pred integer?)
+            (s/optional-key :website) s/Str
+            (s/optional-key :polling-place-url) s/Str
+            (s/optional-key :polling-place-url-shortened) s/Str
+            (s/optional-key :local-voting-methods) #{election-schema/local-voting-method-schema}
+            (s/optional-key :source) election-schema/source-schema})
+   election-schema/check-order!))
```
mrichards42: this needs to be `check-dates` right? not `check-order!`

teaforthecat:
Oh I see, I missed the ballot-received compared to the other ballot-* dates.

mrichards42:
@teaforthecat I'm pretty sure we still want this for checking deadlines against the election date -- are you planning on continuing this in another PR?

teaforthecat:
I like to push buttons

src/election_works/entities/election_schema.clj
```diff
@@ -135,3 +135,52 @@
...
+   [:ballot-return-deadline-received :election/date]])
+
+(defn check-dates
+  "like s/validate, throws an exception if the dates are out of order, or returns
+  the election"
+  [election]
+  (let [date (:date election)
+        dd (:district-divisions election)
+        voting-methods (map :voting-methods dd)
+        voting-registration-methods (map :voting-registration-methods dd)
```
mrichards42: That's `:voter-registration-methods`

src/election_works/entities/election.clj
```diff
@@ -537,10 +537,16 @@
 
 (def read-schema s/Uuid)
 
+(defn validate
+  "return the election, unless one of these checks throws an exception of course"
+  [schema election]
+  (s/validate schema election)
+  (election-schema/check-dates election))
```
mrichards42: I like this a lot -- mind also changing the call here? https://github.com/democracyworks/election-works/blob/355fdbff01ca304628db1f6d66872b946a67559e/dev-src/election_works/import.clj#L269

teaforthecat: done

src/election_works/entities/election_schema.clj
teaforthecat: oh boy, that was a close one. I added a test just to be thorough. Thanks!

PR #105, opened 2019-10-30
Upcoming election update

leilakass:
Changes upcoming elections to include elections on the current day and using a different implementation of time.


dzcode:
Same comments as the other one, but otherwise this looks good

PR #107, opened 2020-01-07
Primary election instructions

solomonsokoya:
[Jira Card](https://test-ve.atlassian.net/browse/VE-39?atlOrigin=eyJpIjoiOWU5ZGU4NTBkMGUxNDljM2I0YWE5MzQ1MTUwMDIwNTEiLCJwIjoiaiJ9)

Election type added to the db schema and throughout the election works service.


dev-src/election_works/import.clj
```diff
@@ -288,8 +288,12 @@
                      :unexpected-columns bad-keys})))
   (-> {:description (row-get csv-row "description")
        :date (row-get-instant csv-row "date")
+       :type (some-> (row-get csv-row "type")
+                     string/lower-case
+                     (string/replace #"\s+" "-")
+                     keyword)
        :district-ocd-ids (some-> (row-get csv-row "district-division")
-                                 (string/split #"\s+"))}
+                                 (string/split #"\s"))}
```
mrichards42: How come this change? I'd actually be inclined to trim all the spaces:
```clj
(some-> (row-get csv-row "district-division")
        (string/trim)
        (string/split #"\s+"))
```

dev-src/election_works/import.clj
```diff
@@ -288,8 +288,12 @@
                      :unexpected-columns bad-keys})))
   (-> {:description (row-get csv-row "description")
        :date (row-get-instant csv-row "date")
+       :type (some-> (row-get csv-row "type")
+                     string/lower-case
+                     (string/replace #"\s+" "-")
+                     keyword)
```
mrichards42: Could we ask the research team to write these out for us with dashes instead of spaces? They're familiar with that sort of syntax (see, e.g. the `source-type` column). I expect that at some point we're going to want `:election/type` to be cardinality many, and it would be a simpler transition if we can use spaces as a delimiter in the spreadsheet.

src/election_works/entities/election.clj
```diff
@@ -578,6 +584,7 @@
 (def create-schema
   {(s/optional-key :id) s/Uuid
    :description s/Str
+   :type s/Keyword
```
mrichards42: Should this be optional?

src/election_works/entities/election.clj
```diff
@@ -1,20 +1,22 @@
...
+   [datomic.api :as d]
+   [datomic-toolbox.core :as db]
+   [datomic-toolbox.transaction :as tx]
+   [datomic-toolbox.query :as query]
+   [com.climate.newrelic.trace :refer [defn-traced]]
+   [election-works.entities.election-schema :as election-schema]
+   [election-works.time :as time]
+   [schema.core :as s])
+  (:import
+   (java.lang IllegalArgumentException)))
```
mrichards42: I know that a certain prominent member of the clojure community [favors this style](https://stuartsierra.com/2016/clojure-how-to-ns.html), but IMO we should at least strive to keep a consistent style within the same repo, and none of the other namespaces in this repo have a line break after `:require` or `:import` (nor does, for that matter, our repo [template](https://github.com/democracyworks/kraken-works-lein-template)).

src/election_works/entities/election.clj
mrichards42: And while we're at it, would it be worth whitelisting allowed types (like we did with `source-type`) to prevent typos that would have unintended consequences down the line?

solomonsokoya: After running through the spreadsheet I've noticed all of the election uploaded has type filled in. I think we should make it a necessary key.  Thoughts? @mrichards42

mrichards42: I think I'd let research or product make the call in both cases actually.

resources/schemas/20191216-add-election-types.edn
```diff
@@ -0,0 +1,5 @@
+[{:db/ident :election/type
+  :db/valueType :db.type/keyword
+  :db/cardinality :db.cardinality/one
+  :db/index true
+  :db/doc "The type of election we're dealing with, e.g., :primary, :general, etc."}]
```
sduckett: 🎏 we'll want to adjust this docstring now that we have more of an idea of what it looks like.

resources/schemas/20191216-add-election-types.edn
```diff
@@ -0,0 +1,5 @@
+[{:db/ident :election/type
+  :db/valueType :db.type/keyword
+  :db/cardinality :db.cardinality/one
+  :db/index true
+  :db/doc "The (for now, single) type of election we're dealing with, e.g., :primary, :general, etc."}]
```
eggsyntax: Can you clarify "for now, single"? Do you mean that this might be a cardinality-many attribute at some point? With Datomic schema you almost NEVER want to change fundamental properties of an attribute, so if it may need to be card-many, we probably need to either a) make it card-many from the start, or b) plan on using a separate attribute later (maybe `:election/types`?).

Or am I just misunderstanding what you're saying in the comment?

src/election_works/entities/election.clj
```diff
@@ -237,7 +239,7 @@
            polling-place-url
            polling-place-url-shortened
            local-voting-methods
-           source]}]
+           source type]}]
```
eggsyntax: Tiny quibble: I'd encourage putting `type` on a separate line so that it doesn't get overlooked since the other lines are single items -- multiple items on the same line are an idiosyncrasy that we inherited, but I think we should strive to improve it where possible. 

PS - After expanding the code I see that there are some earlier lines in this vector that have multiple items, so either way :)

resources/schemas/20191216-add-election-types.edn
eggsyntax: (note -- some discussion of this now in #tv-devs)

src/election_works/entities/election.clj
solomonsokoya: Based on what the team said. It will not be optional and I have added allowed types

resources/schemas/20191216-add-election-types.edn
solomonsokoya: Based on the discussion. My interpretation is leave it single. I'll add the types we are dealing with into the docstring.

dev-src/election_works/import.clj
solomonsokoya: @sduckett refresh my memory. why did we decided to do this?

sduckett: Looking at the commit, I can't remember. Maybe it was a mistake?

solomonsokoya: Seems like it. Removing

src/election_works/entities/election.clj
mrichards42: You would also want to revert the spacing changes for the other lines (you should just be able to run your editors auto-formatter on the ns form).

Also same comment on the other big ns-form change in entities.election-test

sduckett: This was probably me; I also favor this style (and wish our template did too.) It's isolated not to avoid consistency, but to start growing toward that. (and I'm happy to pretend to stop favoring the style when I touch namespaces 'round here.)

sduckett:
> @sduckett and @solomonsokoya I see that most of the commits have some notes in them (`[wip]`, `[squash]`, etc.). Is the idea to squash these all into a single commit, or rebase them into a couple commits?

I believe so. My usual plan is to reword them once review was done, force-push, then rebase and merge (or squash, that's fine too, but I lean rebase.)

src/election_works/entities/election.clj
mrichards42: I'm certainly open to discussion. Most of the tv repos don't have the line break (I also happen to have a slight preference for no line break, but it might just be b/c that's what I'm used to seeing in TV). I don't care all that much, but unless we decide as a team to pick one or the other, we're just going to have devs swapping back and forth all the time, and IMO the extra diff is distracting.

sduckett: fine by me, as long as the namespaces are sorted ;)

mrichards42:
> My usual plan is to reword them once review was done, force-push, then rebase and merge (or squash, that's fine too, but I lean rebase.)

Just a slight preference for cleaning up WIP commits before it lands on master, but I don't care how it happens (and of course sometimes I forget and hit "merge" on a branch I meant to clean up 🙄).

src/election_works/entities/election.clj
mrichards42: amen to that!

eggsyntax: @sduckett what do you like about that style? I like the more common convention that Mike suggested because it makes it much easier to see where the different sections start (in particular it's useful to know at a glance if there are `:import`s happening). I don't find Stuart's argument on that at all convincing because we read frequently but sort rarely, so it doesn't make sense IMHO to prioritize ease-of-sorting. Also it seems inconsistent to me; if it was truly worth prioritizing for `sort-lines` you'd have to put the closing paren on a separate line too.

sduckett: @eggsyntax Mike asked me the same question and here's what I said:

Screen Shot 2020-01-08 at 13 52 50

I don't make the `sort-lines` argument, I just like not-wide forms in code.

eggsyntax: Interesting. In TV, at least, that hasn't seemed like a problem (and I often split my editor into three panes vertically, so I'm pretty sensitive to line length. Looking at buffers I have open in three different projects, the longest thing I'm seeing is 
```
(ns election-works.entities.election
  (:require ...
            [election-works.entities.election-schema :as election-schema]))
```
which still tops out at 74 chars.

(I'm sure there are longer ones somewhere in TV, it just seems like it's almost never a problem)

PR #108, opened 2020-03-13
HTV-87 add fields to make imports work

teaforthecat:


src/election_works/entities/election_schema.clj
```diff
@@ -56,6 +64,7 @@
     :excuse-required                                    (s/eq false)
     (s/optional-key :note                             ) s/Str
     (s/optional-key :instructions                     ) instructions-schema
+    (s/optional-key :ballot-request-form-url)           i18n-map-uri-schema
```
dzcode: Nit: can you fix the parens spacing to match the other lines here?

src/election_works/entities/election_schema.clj
```diff
@@ -5,6 +5,8 @@
 
 (def authority-level-schema (s/enum :state :county :municipal))
 
+(def supported-languages #{:en :es})
```
mrichards42: IMO it would be worth making this `(s/enum :en :es)`, and then also replacing the literal version of that in `instructions-schema` with `supported-languages`. I'm honestly not suport familiar with schema, but it looks like we're using `s/enum` everywhere else in this file for this kind of thing.

src/election_works/entities/election_schema.clj
teaforthecat: I'd be happy to fix the other ones if you like, those look silly to me.

src/election_works/entities/election_schema.clj
teaforthecat: done

src/election_works/entities/election_schema.clj
teaforthecat: they look similar now - done

teaforthecat:
@dzcode These are the fields that broke the election import. I thought I was going to have to add more too, but it isn't clear to me that they have to be added to the election here. I'm tracing how the context is being built now, but if you have any input that would be appreciated. Maybe we can work together to get the alias in the context if it isn't already.

mrichards42:
@teaforthecat we definitely need to add `:alias` to voting methods, that's the test I ran that failed over in https://github.com/democracyworks/election-authority-works/pull/70#issuecomment-588282458

party-instructions are on the authority root only, since they don't belong to a specific voting method or registration method, so it's ok to leave them off the election IMO.

teaforthecat:
huh, not sure why I didn't see that in my testing locally

teaforthecat:
Sorry I could use some help, I'm not sure what else needs to be done here.

teaforthecat:
I see that now. I won't be able to finish this today, sorry. Talk to you on Monday.

teaforthecat:
I think this is done now. The data is being stored to the db (I was thinking it was merged in later for some reason). There is a neat test that looks like it proves that the new fields are being stored and retrieved. (That test is neat but if there are any failures it is a nightmare.) I'm hoping the `test-election` shows where the fields are but if there is somewhere else I could add them, I could do that.

teaforthecat:
I've tested this with various conditions and it looks good, I'll go ahead with merging.

PR #109, opened 2020-03-24
Add early-voting-varies attribute

eggsyntax:
[NOTE - won't be merged until that's been decided by product/research]

@dzcode tagging you as reviewer since you've just been looking at related code, but if you have too much on your plate let me know & I'll tag someone in.

@thackley tagging you since IIRC we looked at this a bit.


dev-src/election_works/import.clj
```diff
@@ -145,7 +154,10 @@
                  (search-authorities {:ocd-divisions [ocd-division]
                                       :language :all}))]
     (if (and results (= :ok (:status results)))
-      (first (:authorities results))
+      (if (seq (:authorities results))
+        (first (:authorities results))
+        (throw (RuntimeException.
+                (str "No authorities found matching " ocd-division))))
```
eggsyntax: Unrelated to adding this attribute; just a minor improvement to the import process.

dzcode: Nit: I think this could be
```
(if-let [authorities (seq (:authorities results))]
  (first authorities)
  (throw (RuntimeException. ...
```
Very take it or leave it, though 🤷‍♂

dzcode: Also thank you for adding this!

mrichards42:
> I think you would also want to add this to election-authority-works, and have the election-level field here be an override.

That's probably a question for research or product -- is this the kind of attribute that only makes sense on an election, or is there a reasonable state default we should include in election-authority-works?

On that note -- you've added this as a top-level attribute on the election (`:election/early-voting-varies`). IMO this would make more sense as an attribute on the early-voting method (`:voting-method.early-voting/varies`). If you disagree, could you talk a little bit about why it belongs on the election itself instead of on early-voting?

eggsyntax:
Carmen confirms that it's election-only, not applicable to the authorities.

(will have to get back to Mike's comment later)

eggsyntax:
> On that note -- you've added this as a top-level attribute on the election (`:election/early-voting-varies`). IMO this would make more sense as an attribute on the early-voting method (`:voting-method.early-voting/varies`). If you disagree, could you talk a little bit about why it belongs on the election itself instead of on early-voting?

My thinking was twofold, I guess: first, that it was framed as a property of elections. While it's not really a product-level decision, Carmen's take on it is "I mean, it is kinda about the voting method in that it is connected to early-voting, but we don’t treat early voting the same way we do :by-mail or :in-person." Second, that I kind of hate to give entities an attribute that's meaningless for most instances of the entity.

That said, now that I've dug a little deeper, I see that we're doing that already with `:voting-method/url`, which is only applicable to early voting (and `:voting-method/start` / `:voting-method/end`, which are only applicable to early-voting and in-person). That makes me think you're probably right that it's a better approach.

eggsyntax:
Replaced by #110

PR #110, opened 2020-03-27
[TVF-276] Early voting varies vm

eggsyntax:
Rewrite of PR #109, putting the new attribute on the (early-voting) voting-method as MR suggested, and also reversing the meaning of the attribute (from `early-vote-varies` to `exact-early-voting-dates` [`voting-method.early-voting/exact-dates` in datomic]) for clarity.

Note the docstring re: csv having to be well-formed -- turns out blank lines break it in a confusing way. We could
- ignore it because it's not a common problem.
- rewrite to use `slurp` instead of `io/reader` so that we can easily filter out blank lines before parsing (at some potential cost to efficiency).
- move away from the ~`parse-csv`~ [`csv-map`](https://github.com/mihi-tr/csv-map) lib which assumes well-formedness in favor of another csv-parsing lib, maybe [`clojure-csv`](https://github.com/davidsantiago/clojure-csv), the lib underlying `csv-map`.

Interested to hear thoughts on that decision. I'll probably create a small-fix ticket unless we choose option 1.


eggsyntax:
[Note: still have to update tests later this afternoon]

dev-src/election_works/import.clj
```diff
@@ -372,7 +391,7 @@
   (partial
    read-csv
    (fn [csv-row]
-     (print "Reading election" (row-get csv-row "description") "...")
+     (println "Reading election" (row-get csv-row "description") "...")
```
mrichards42: The lack of a newline was on purpose, so the script would print:
```
Reading election Some Test Election...OK
```
rather than
```
Reading election Some Test Election...
OK
```

dev-src/election_works/import.clj
```diff
@@ -24,6 +24,7 @@
     "early-voting-start"
     "early-voting-end"
     "early-voting-url"
+    "exact-early-voting-dates"
```
mrichards42: Nit: I might go with `early-voting-exact-dates` for consistency w/ the other early-voting fields, but not a big deal either way.

README.md
```diff
@@ -163,6 +165,18 @@ The CSV file should have the following headers (value hints in parens). Order un
...
 * excuse-required (TRUE of FALSE; whether the non-primary by-mail voting method requires an excuse)
 
+### Adding a column to the spreadsheet
+
+In order to add a column to the spreadsheet, you'll need to (at least)
+1. Discuss it with someone from the research department (who will add it to the actual spreadsheet
+once the code is done).
+1. Add the attribute to the DB schema.
+1. Handle it in the `election-works.import` script.
+1. Add it to the election schemas in `election-works.entities.election`
```
mrichards42: I'd also add a line about potentially adding it to election-authority-works (if it's a voting-method or reg-method field)

dev-src/election_works/import.clj
```diff
@@ -358,6 +374,9 @@
          (maybe-assoc :local-voting-methods local-voting-methods)))))
 
 (defn read-csv
+  "Reads a file matching the csv export of the DW elections spreadsheet. csv
+  file MUST be well-formed (eg blank lines will cause problems)."
+  ;; TODO discuss in PR
```
mrichards42: I'd probably go with (1) and say it isn't usually a problem. But you're welcome to do whatever you'd like with it if it bugs you.

dev-src/election_works/import.clj
eggsyntax: Oh, right. Cool, I'll change that back.

dev-src/election_works/import.clj
eggsyntax: The failure mode's pretty awful -- for every blank line you get a map like `{:name-of-rightmost-header ""}`. Lost a painful amount of time to that, since that was the header I'd just added and I assumed it was a problem with my code rather than the lib ;P

eggsyntax:
> Can we have question marks in column names? If so, I think the meaning would be way more clear with one in this case

Works fine at the code level; I'll double-check w/ product & research as well.

mrichards42:
> Can we have question marks in column names? If so, I think the meaning would be way more clear with one in this case

Agreed with egg that it's probably fine code-wise, but we haven't used question marks for any other boolean fields. IMO if we want a question mark for this field we should also use it for the other booleans.

dzcode:
It might be good to fix the others too where they're less clear, but I don't think that needs to happen in this PR since this needs to get out (if we go with the question mark)

eggsyntax:
:man_shrugging: research prefers no question mark. And now that I ask them about it, they may want to swap it back to `early-voting-varies` (or similar) rather than `early-voting-exact-dates`. They'll get back to me later today on it. I still think the latter is clearer; I'm a bit torn on whether (if they say they want to swap it back) to keep the attribute as the version that I think is clearer, or to switch it back to have it be consistent with the spreadsheet.

teaforthecat:
imho, I like the "-varies" version better because in the template, I think it will read better. Something like
``` (if early-voting-varies show-early-voting-varies-sentence)```

mrichards42:
IMO a big reason to go with `exact-dates` is that we've been treating elections as if none of them have exact early voting dates (i.e. all of them vary) so far. If we go with `exact-dates`, our existing elections won't have that value, so our code would treat them as if they were `exact-dates false` by default -- which is what we've been doing all along. Whereas if we went with `varies`, we'd have to then change every existing election to be `varies true`, since otherwise we'd pick the wrong code path by default.

eggsyntax:
Just heard back from research; they're fine with going with `exact-dates` :sweat_smile:

PR #112, opened 2020-04-21
Validation supported voting methods

thackley:
Validating that election methods are supported by the authority for [TVF-331](https://democracyworks.atlassian.net/browse/TVF-331)


src/election_works/voting_methods.clj
```diff
@@ -17,12 +18,19 @@
...
+  "Keep only the authority voting method types that are specified by the election's voting methods. Defaults to the full set of authority voting methods if election voting methods are not specified."
+  [election-supported-methods authority-voting-methods]
+  (if election-supported-methods
+      (if-let  [unknown-methods (seq (set/difference (set election-supported-methods) (set (map :type authority-voting-methods))))]
+        (throw (ex-info (str "Key(s) " unknown-methods " found in election's supported voting methods but not supported by authority.")
+                        {:unknown-methods unknown-methods
+                         :voting-methods authority-voting-methods
+                         :supported-voting-methods election-supported-methods}))
+        (filter #(contains? (set election-supported-methods) (:type %)) authority-voting-methods)
+        )
```
dzcode, line 31: Nit: Different from other languages, the style standard in clojure is to put all closing parentheses/brackets/etc. on the line above, so this would be:
```
                        {:unknown-methods unknown-methods
                         :voting-methods authority-voting-methods
                         :supported-voting-methods election-supported-methods}))
        (filter #(contains? (set election-supported-methods) (:type %)) authority-voting-methods))
    authority-voting-methods))
```

Same thing at the end of the `test/election_works/voting_methods_test.clj` file

src/election_works/voting_methods.clj
```diff
@@ -17,12 +18,19 @@
...
-  "Keep only the supported voting-method types."
-  [supported-methods voting-methods]
-  (if supported-methods
-    (filter #(contains? supported-methods (:type %)) voting-methods)
-    voting-methods))
+(defn keep-election-supported-voting-methods
+  "Keep only the authority voting method types that are specified by the election's voting methods. Defaults to the full set of authority voting methods if election voting methods are not specified."
+  [election-supported-methods authority-voting-methods]
+  (if election-supported-methods
+      (if-let  [unknown-methods (seq (set/difference (set election-supported-methods) (set (map :type authority-voting-methods))))]
```
dzcode, line 25: Nit: We don't have a strict standard for how long lines should be, but in general I try to break them up if they're over 100ish characters, which I think this is (you can set a line in your editor to show this). For this line, I'd probably put this `(set (map :type authority-voting-methods))))]` on the next line and then run an auto-formatter to do the indentation correctly

dev-src/election_works/import.clj
```diff
@@ -326,7 +326,7 @@
       (maybe-assoc :reg-deadline-type (row->registration-deadline-type csv-row))
       (maybe-assoc :election-day-registration (election-day-registration? csv-row))
       (maybe-assoc :primary-voting-method (edn/read-string (row-get csv-row "primary-voting-method")))
-      (maybe-assoc :supported-voting-methods (row->supported-voting-methods csv-row))
+      (maybe-assoc :election-supported-voting-methods (row->election-supported-voting-methods csv-row))
```
dzcode, line 329: I think you can leave this naming alone, since this whole file just operates on elections and the field is just `supported-voting-methods` in the spreadsheet. Personally I like the extra clarification in the `keep-election-supported-voting-methods` function because we have two different sets of voting methods, but elsewhere I think it's clear enough (feel free to disagree though).

src/election_works/voting_methods.clj
```diff
@@ -77,8 +85,8 @@
   "Create the voting-methods for the election by combining
   the voting-methods of the relevant state or election authorities
   and merging them with the election-specific properties"
-  [authority-vms election-vms {:keys [supported-voting-methods primary-voting-method]}]
-  (some-> (keep-supported-voting-methods supported-voting-methods authority-vms)
+  [authority-vms election-vms {:keys [election-supported-voting-methods primary-voting-method]}]
+  (some-> (keep-election-supported-voting-methods election-supported-voting-methods authority-vms)
```
dzcode, line 89: Per my comment above, I'd keep the function name changed here, but change the `election-supported-voting-methods` symbol back to just `supported-voting-methods`

src/election_works/voting_methods.clj
mrichards42: ➕ I aim for 80, and allow 100 if it's really inconvenient to break a line, though like dan said, there's no strict rule.

dev-src/election_works/import.clj
eggsyntax: I think I advocated for the increased clarity, just because it's easy to confuse the voting-methods for the election with the voting-methods for the authority. Less of an issue in this file, though, and I think either way is fine.

thackley: For me, both as someone new to the code and just me a person in general, having the additional clarity of `election`, especially since I was simultaneously having to think about election authorities and elections. I also think having it as `election-supported-voting-methods` in the `->voting_methods` function provides some additional clarity. 

That said, I can see how not having it there aligns better with the spreadsheet and, given the context of being within `election-works` it could be unnecessary. I can easily go either way and am happy to go with the consensus/preferences of a group.

dzcode: That makes sense, we can leave it as it as you have it then 👍

test/election_works/voting_methods_test.clj
```diff
@@ -133,3 +133,18 @@
...
               merged-vms))
       (s/validate #{voting-method-schema} merged-vms))))
+
+(deftest keep-election-supported-voting-methods-test
+  (testing "keep supported voting methods fix this later"
+    (let [authority-vms #{{:type :by-mail, :primary false, :excuse-required true,
+                           :instructions {:voting-id {:en "Fake CO by-mail voting ID instructions"}}}
+                          {:type :in-person, :primary true, :excuse-required false,
+                           :instructions {:voting-id {:en "Fake CO voting ID instructions"}}}}
+          ]
```
dzcode, line 143: Nit: Parentheses spacing is off here as well

